PR #18, opened 2022-08-29
pa_calibrate.cfg : specify first band # , fix M117

fbeauKmi:
Hi,
here is a small fix for pa_calibrate.

### DIFF
--- a/macros/calibration/pa_calibrate.cfg
+++ b/macros/calibration/pa_calibrate.cfg
@@ -132,6 +132,7 @@ gcode:
       "Your final Pressure Advance value would then be:"
     )}
     {action_respond_info("")}
+    {action_respond_info("Assuming that first band_number is \0430")}
     {action_respond_info("PA = %.4f + (%.4f * band_number)" % (pa_start, pa_increment))}
     {action_respond_info("")}
     {action_respond_info("Update your [extruder] config section and set \"pressure_advance\" to the value you calculated.")}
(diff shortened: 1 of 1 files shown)


macros/calibration/pa_calibrate.cfg
```diff
@@ -132,6 +132,7 @@ gcode:
       "Your final Pressure Advance value would then be:"
     )}
     {action_respond_info("")}
+    {action_respond_info("Assuming that first band_number is \0430")}
```
Frix-x, line 135: What is this \0430 ?

Frix-x (changes requested):

macros/calibration/pa_calibrate.cfg
fbeauKmi: I should avoid that. \043 is octal value of unicode character # which makes in console "Assuming that first band_number is #0". If you consider the first band_number is 1 then the equation `PA = pa_start + ( pa_increment * band_number)` is wrong. 

Actually, you can consider two ways: Either you print the first band  with `Pressure_advance = pa_start` then first band_number = 0 or you print with   `Pressure_advance = pa_start  + pa_increment` then first band_number will be 1.

Frix-x (approved):
Ok for me. It's better to keep it simple and readable :)

macros/calibration/pa_calibrate.cfg
Frix-x: It's ok, thank for the explanation about the unicode char, I was not aware this was possible !

PR #20, opened 2022-08-31
add-option-in-start-end-print

elpopo-eng:
# To always make a full homing and QGL in start print
`variable_force_homing_in_start_print: False`

# To disable motors in end print
`variable_disable_motors_in_end_print: True`


Frix-x (approved):
Ok for me with these new variables !

PR #24, opened 2022-10-01
Small macro adjustments

elpopo-eng:
A lot of minor adjustments :

- Remove some M400
- Add lift_z for PARK macro
- Enhance filament swap
- adjustment in start print

### DIFF
--- a/config/software/bed_mesh.cfg
+++ b/config/software/bed_mesh.cfg
@@ -1,6 +1,6 @@
 [bed_mesh]
 # FOLLOW THE FORUMULA FOR relative_reference_index to be centered, 
-# This is only for manual call of BED_MESH_CABLIBRATE all those parameters can overridden on the fly if using the adaptive bed mesh macro
+# This is only for manual call of BED_MESH_CABLIBRATE all those parameters are overridden on the fly if using the adaptive bed mesh macro
 # for example:
 # [(9x9)-1] / 2 = 40
 # [(7x7)-1] / 2 = 24
--- a/config/software/bed_mesh.cfg
+++ b/config/software/bed_mesh.cfg
@@ -1,5 +1,7 @@
 [bed_mesh]
-# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, for example:
+# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, 
+# This is only for manual call of BED_MESH_CABLIBRATE all those parameters can overridden on the fly if using the adaptive bed mesh macro
+# for example:
 # [(9x9)-1] / 2 = 40
 # [(7x7)-1] / 2 = 24
 # [(5x5)-1] / 2 = 12
--- a/config/software/bed_mesh.cfg
+++ b/config/software/bed_mesh.cfg
@@ -1,4 +1,8 @@
 [bed_mesh]
+# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, for example:
+# [(9x9)-1] / 2 = 40
+# [(7x7)-1] / 2 = 24
+# [(5x5)-1] / 2 = 12
 speed: 350
 horizontal_move_z: 12
 mesh_min: 10, 25
--- a/config/software/z_calibration.cfg
+++ b/config/software/z_calibration.cfg
@@ -8,5 +8,13 @@ switch_xy_position: 198.5,285.5
 switch_offset: 0.390
 max_deviation: 0.750
 samples: 3
+#samples_tolerance: 0.01
+#clearance: 8
 speed: 350 # X,Y movements
 probing_first_fast: true
+#probing_speed: 20
+#probing_second_speed: 5
+#probing_retract_dist: 2
+start_gcode: ATTACH_PROBE
+#before_switch_gcode:
+end_gcode: DOCK_PROBE
--- a/macros/helpers/nozzle_cleaning.cfg
+++ b/macros/helpers/nozzle_cleaning.cfg
@@ -51,15 +51,7 @@ gcode:
 
     G90
 
-    {% if purge_and_brush_enabled %}
-        # Move to purge zone only if it's available, else just purge whe the toolhead is
-        {% set Px, Py, Pz = printer["gcode_macro _USER_VARIABLES"].purge_bucket_xyz|map('float') %}
-        {% if Z_DROP == 1 %}
-            G1 X{Px} Y{Py} Z{Pz} F{St}
-        {% else %}
-            G1 X{Px} Y{Py} F{St}
-        {% endif %}
-    {% endif %}
+    MOVE_TO_PURGE_BUCKET
 
     # Heat if needed and purge
     _LOW_TEMP_CHECK T={TEMP}
@@ -87,12 +79,12 @@ gcode:
     {% set purge_and_brush_enabled = printer["gcode_macro _USER_VARIABLES"].purge_and_brush_enabled %}
     {% set Z_DROP = params.Z_DROP|default(1)|int %}
     
+    # Move to purge zone only if it's available, else just purge where the toolhead is
     {% if purge_and_brush_enabled %}
         SAVE_GCODE_STATE NAME=MOVE_TO_PURGE_BUCKET
         G90
         {% set St = printer["gcode_macro _USER_VARIABLES"].travel_speed * 60 %}
 
-        # Move to purge zone only if it's available, else just purge whe the toolhead is
         {% set Px, Py, Pz = printer["gcode_macro _USER_VARIABLES"].purge_bucket_xyz|map('float') %}
         {% if Z_DROP == 1 %}
             G1 X{Px} Y{Py} Z{Pz} F{St}
@@ -101,4 +93,6 @@ gcode:
         {% endif %}
         RESTORE_GCODE_STATE NAME=MOVE_TO_PURGE_BUCKET
     {% endif %}
-    
+
+       
+ 
--- a/macros/helpers/nozzle_cleaning.cfg
+++ b/macros/helpers/nozzle_cleaning.cfg
@@ -79,3 +79,26 @@ gcode:
     G4 P{20 * 1000}
   
     G92 E0
+
+
+[gcode_macro MOVE_TO_PURGE_BUCKET]
+description: Move over the purge bucket
+gcode:
+    {% set purge_and_brush_enabled = printer["gcode_macro _USER_VARIABLES"].purge_and_brush_enabled %}
+    {% set Z_DROP = params.Z_DROP|default(1)|int %}
+    
+    {% if purge_and_brush_enabled %}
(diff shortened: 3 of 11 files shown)


config/software/bed_mesh.cfg
```diff
@@ -1,4 +1,8 @@
 [bed_mesh]
+# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, for example:
```
Frix-x, line 2: Can you also add a comment saying that this is the default config when calling `BED_MESH_CALIBRATE` without any parameters but it will be overiden on the fly if using the adaptive bed mesh macro ?

macros/helpers/nozzle_cleaning.cfg
```diff
@@ -79,3 +79,26 @@ gcode:
...
+        # Move to purge zone only if it's available, else just purge whe the toolhead is
+        {% set Px, Py, Pz = printer["gcode_macro _USER_VARIABLES"].purge_bucket_xyz|map('float') %}
+        {% if Z_DROP == 1 %}
+            G1 X{Px} Y{Py} Z{Pz} F{St}
+        {% else %}
+            G1 X{Px} Y{Py} F{St}
+        {% endif %}
+        RESTORE_GCODE_STATE NAME=MOVE_TO_PURGE_BUCKET
+    {% endif %}
+
```
Frix-x, line 104: Please add a newline at the end of the file :)

macros/helpers/nozzle_cleaning.cfg
```diff
@@ -79,3 +79,26 @@ gcode:
...
+gcode:
+    {% set purge_and_brush_enabled = printer["gcode_macro _USER_VARIABLES"].purge_and_brush_enabled %}
+    {% set Z_DROP = params.Z_DROP|default(1)|int %}
+    
+    {% if purge_and_brush_enabled %}
+        SAVE_GCODE_STATE NAME=MOVE_TO_PURGE_BUCKET
+        G90
+        {% set St = printer["gcode_macro _USER_VARIABLES"].travel_speed * 60 %}
+
+        # Move to purge zone only if it's available, else just purge whe the toolhead is
```
Frix-x, line 95: where* the toolhead is

macros/print_base/park.cfg
```diff
@@ -4,24 +4,30 @@ gcode:
...
     {% endif %}
+    G90
+    G1 Z{z_safe} F{Sz}
 
+    {% if printer.extruder.temperature > 185 %}
     G91    
-    G1 Z{z_safe} F{Sz}
+        G1 E-{E} F2100
     G90
+    {% endif %}
```
Frix-x, line 31: I'm OK to use G10/G11 here if you want.

But then, why having then another E-{E} ? I suspect it's for the case where we do PARK E=10 like at the end of a print ?
Then better to change the default E value to 0 in this case. So if no E parameter is put in the call, then there will be no "double retract"

macros/print_base/start_print.cfg
```diff
@@ -112,6 +114,8 @@ gcode:
         RESPOND MSG="Extruder temperature OK"
     {% endif %}
 
+    # perform homing after hotend heating
+    G28 Z
```
Frix-x, line 118: Why not anymore after the nozzle clean ?
I think it's better to do it as late as possible to give it as much time as possible to stabilize. Also, I think it's better to do it after a nozzle clean as it's more stable without any remaining filament. Don't you think ?

macros/probing/dockable_probe.cfg
```diff
@@ -14,13 +14,11 @@ gcode:
     {% set function  = 'pre_' ~ params.FUNCTION %}
     {% set move  = params.MOVE|default(0) %}
     {% set speed  = params.MOVE_SPEED|default(0) %}
-    M400 # mandatory to save the new safe position
     RESTORE_GCODE_STATE NAME={function} MOVE={move} MOVE_SPEED={speed}
 
 [gcode_macro _ENTRY_POINT]
 gcode:
     {% set function  = 'pre_' ~ params.FUNCTION %}
-    M400 # mandatory to save the new safe position
```
Frix-x, line 23: These two `M400` lines are really mandatory for slower Pi to be sure that the movement is finished and the buffer flushed before doing a save of the position.

FYI, these M400 lines in the official Klicky repo are also written by myself as I corrected a bug in the early klicky macros developpement, so that's why I know exactly why they are here :sweat_smile:

config/software/z_calibration.cfg
```diff
@@ -8,5 +8,13 @@ switch_xy_position: 198.5,285.5
...
 max_deviation: 0.750
 samples: 3
+#samples_tolerance: 0.01
+#clearance: 8
 speed: 350 # X,Y movements
 probing_first_fast: true
+#probing_speed: 20
+#probing_second_speed: 5
+#probing_retract_dist: 2
+start_gcode: ATTACH_PROBE
```
Frix-x, line 18: Is this working OK ?
As you did not make any change to the macros to remove the call to `ATTACH/DOCK`, so maybe there is some place where it's not needed anymore if we keep it here ?

Frix-x (commented):
Thank you for this PR !

I think most of it is OK for me to merge. Just a couple of questions to be sure I correctly understood your needs and changes.

macros/probing/dockable_probe.cfg
elpopo-eng: I agree with that, but we need to save the position only in 2 cases:
- PROBE_CALIBRATE
- PROBE_ACCURACY

In all other case the previous position is not used.

If you look these 2 macros i have insert a M400 before calling ENTRY_POINT

config/software/z_calibration.cfg
elpopo-eng: This is for docking probe after a probing error.
(updated in v0.9.2 of plugin Z-calibration: _The probe is now docked again before raising an error, if end_gcode is used (thanks to top-gun)_

I'm agree as we lock probe in start_print this may occured only when we are working in manual

macros/print_base/start_print.cfg
Frix-x: Wow, is that already happened to you ?
But Ok, then we add this homing after the heating of the nozzle.

But maybe we also need to add another G28Z after the clean nozzle ? If we don't do that, we will have a lot of different Z offsets computed by the Z calibration plugin as the offset will be applied depending of how much filament there was on the nozzle I think...

macros/probing/dockable_probe.cfg
Frix-x: Ok for me then ! It doesn't change a lot of things afterall :)

config/software/z_calibration.cfg
Frix-x: Mhmm OK. It's a good thing indeed to dock the probe in case of an error !
Ok, if you tested it and it's working correctly (I mean no double attach move or double detach move), we can add it now. But I'll try to investigate it more in deep to understand every possible case and do it properly.

macros/print_base/park.cfg
Frix-x: Ok, I've checked all the file. As we are also doing a manual E-10 in end_print and cancel_print, then we can remove it from here as you suggested. Can you do it ?

macros/print_base/start_print.cfg
elpopo-eng: Yes it's arrived with a 0.8mm nozzle, i had a fat cold blob of plastic on the nozzle.

It's not necessary to add G28 Z after cleaning, we will be a little higher but it's not important, z-calibration don't take care of it

config/software/bed_mesh.cfg
elpopo-eng: Done

macros/helpers/nozzle_cleaning.cfg
elpopo-eng: Done

macros/helpers/nozzle_cleaning.cfg
elpopo-eng: I hope it done, because github desktop suppress them sometimes....

macros/print_base/park.cfg
elpopo-eng: > Ok, I've checked all the file. As we are also doing a manual E-10 in end_print and cancel_print, then we can remove it from here as you suggested. Can you do it ?

These retract are necessary to be able to manually remove the filament at the end/cancel print.
Might be a good idea to put that 10mm retract in _TIP_Shaping, right?

Frix-x: Mhmm not a really fan of that as we could also just call `_TIP_SHAPING` for something else like for the ERCF or I don't know. That means the `_TIP_SHAPING` could not be associated everytime with a filament unload I think.

macros/print_base/start_print.cfg
Frix-x: Mhmm Ok let's try it and if I have trouble with the offset (I like to get consistent offsets between prints to check everything is alright just looking at the number), I'll add another home Z.

macros/print_base/park.cfg
elpopo-eng: Ok, so we need to keep `G0 E-10` in end/cancel print else we can't remove manually filament.

macros/helpers/nozzle_cleaning.cfg
```diff
@@ -51,15 +51,7 @@ gcode:
...
-    {% if purge_and_brush_enabled %}
-        # Move to purge zone only if it's available, else just purge whe the toolhead is
-        {% set Px, Py, Pz = printer["gcode_macro _USER_VARIABLES"].purge_bucket_xyz|map('float') %}
-        {% if Z_DROP == 1 %}
-            G1 X{Px} Y{Py} Z{Pz} F{St}
-        {% else %}
-            G1 X{Px} Y{Py} F{St}
-        {% endif %}
-    {% endif %}
+    MOVE_TO_PURGE_BUCKET
```
Frix-x, line 54: Be careful to also pass the Z_DROP parameter as I use it when doing an M600 to purge after changing the filament while keeping the Z at the current Z (without falling down at a small Z risking to hit something: this is the goal of the optional Z_DROP)

Frix-x (changes requested):
I added a couple of comments.
But most important one is about Z_DROP

config/software/bed_mesh.cfg
```diff
@@ -1,4 +1,10 @@
 [bed_mesh]
+# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, 
+# This is only for manual call of BED_MESH_CABLIBRATE all those parameters can overridden on the fly if using the adaptive bed mesh macro
```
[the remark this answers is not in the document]
elpopo-eng: Done

macros/helpers/nozzle_cleaning.cfg
elpopo-eng: parameter added

Frix-x (approved):

PR #26, opened 2022-10-09
Light Intensity Variables for print start, printing, and print end

d5aint:
Here is the light intensity variables.

Sorry about closing the original, but was easier to just start over and do it correctly.


Frix-x (approved):
Hey thank you for this new PR.
Look likes it's ok for me now, I'll squash and merge it !

PR #28, opened 2022-10-10
Config v2.0 with optional ERCF support

elpopo-eng:
this PR is an idea of answer to the question: How to include an optional module in the configuration?
I add 2 folders : 
- one in harware 
- one other in macro.

To activate module just uncomment the line in main printer.cfg

At the beginning of each hardware configuration, I added this to add the enable variable in `_USER_VARIABLES`:
```
[gcode_macro _USER_VARIABLES]
# motion_sensor installed in the machine
variable_motion_sensor_enabled: True
gcode:

```


Frix-x:
Before the merge, I will add some clean documentation on the main README.µ
In the meantime if someone wants to push some boards_pins or some config files (like the extruders, etc...), I'll be happy to add them as well :)

fbeauKmi:
As I don't know how to push board pins in this branch. let me put it here.
I have also some comments on hardware config. 
A point is the prefix chip_name. A solution could be to add 2 folders in hardware: dual_mcu & single_mcu in which you can store base and addons specific for both config and keep base and addons for common points. Eventually a third folder for Extruder Can board . 
I added missing pins (CONTROLLER_FAN, RUNOUT_SENSOR, ...)  and correct STATUS_NEOPIXEL=1.24. Filter_fan and exhaust_fan share the same pin. Should exhaust_fan become optional (with a choice filter or exhaust)  
About ligths, as many uses neopixels as caselight I suggest to add a neopixel_caselight.cfg
I also can push lgx_lite.cfg, cw1.cfg and cw2.cfg (You still does not correct gear_ratio on LGX :-) ).

I still can't figure out if is relevant to add mcu*.cfg as include of printer_definition.cfg
[board_pins.zip](https://github.com/Frix-x/klipper-voron-V2/files/9829475/board_pins.zip)

Frix-x:
I also moved in the last commit the [mcu] section with the MCU URI into the board_pins files to have everything in the same place.
Do it also make sense to you ?

fbeauKmi:
Maybe, It is a good option. Consider "mcu" as folder name.
What do you think to add  png MCU's maps in the folder to help wiring, as you don't use Voron standard wiring ? If you are OK,  I could prepare that this week the four board_pins.

Frix-x:
Thanks for the suggestion, I'll have renamed the folder and corrected the typo. Feel free to add the pinouts in a PR on this branch if you want, it's a good idea

Frix-x:
I've made some force-push on this branch to squash my commits. For those who are also working on it, you'll be in a detached head state and you will need to pull again.

Frix-x:
I think we are near the release !
I just need to add the correct cfg files for the standard v2.4 Z axis and also for the Trident Z axis.

Do you see anything else ?

Frix-x:
Now I'm ready to merge ! Is it ok for everyone now ?

PR #25, opened 2022-10-05
Enable vibration calibration in other axis

theninj4:
Thanks for this awesome script and macro! I've been using it to fix some VFA on my v0.1 - I added another input parameter to let me specify different movement patterns for measuring either AB, XY, ABXY, A, B, X or Y.

You're welcome to take this or leave it, or it might be helpful to someone passing through.

--- 

Here are my results demonstrating the value. The A motion resonates at a different frequency to B, X and Y. Tomorrow I'll be checking all the bearings in that belt path 😉 

## A-axis
vibrations_20221005_234305

## B-axis
vibrations_20221005_235447

## X-Axis
vibrations_20221006_000230

## Y-Axis
vibrations_20221006_000937

## ABXY
vibrations_20221006_002545

PR #29, opened 2022-10-10
Theninj4 vibration calibration idea

Frix-x:
Here is a new PR to add the feature drafted by Theninj4 in #25 .

This enable the vibration measurement to be done on different axis like XY, AB, ABXY, A, B, X, Y or Z.
Also some tweaks were made to the Python and Bash scripts for a better formatting of the graphs and to print the movement name in a global graph title.

Here is an example (with no real data) of the result graph new format:

vibrations_20221010_231751


Frix-x:
@theninj4 what do you think about it now ? Can you test it and confirm it works for you before merging ?

elpopo-eng (approved):

PR #31, opened 2022-10-19
Move hard coded defaults to variables.cfg

d5aint:
I have moved the defaults for calibration to the variables.cfg file.
Let me know if this works for you.

### DIFF
--- a/config/variables.cfg
+++ b/config/variables.cfg
@@ -68,7 +68,7 @@ variable_print_default_extruder_temp: 240
 variable_print_default_chamber_temp: 0
 variable_print_default_chamber_max_heating_time: 15
 variable_print_default_soak: 8
-variable_print_default_material: "XXX"
+variable_print_default_material: "ABS"
 
 # Prime line start position
 variable_prime_line_xy: 2.5, 20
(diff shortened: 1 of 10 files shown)


Frix-x:
Thank you for this PR !
As you may have already seen, I have a big one that is currently being cooked (#28 ). It's a big refactoring of the whole folder structure and config files, so I will keep this one on the side for the moment to insert it when the other is merged.

d5aint:
Ok, sounds good.

Frix-x:
Hey, thanks for your patience and also for the sponsoring tips that you sent me ! It's very kind of you !

In the meantime, now that the big merge is done, I'm back to work on this PR. I've merged your changes and refactored it a little bit to this working branch: https://github.com/Frix-x/klipper-voron-V2/tree/calibration_varfile .
Please tell me what do you think about it ?

During this refactoring I've been also thinking that I want to keep the calibration files working by their own. That way if someone want to use them, all he need to do is copy one of the `calibrate_xxx.cfg` file into his own config. Now with this modification, they have two file to copy (the macro and the variable file): what do you think about moving the `{% set .... %}` into the helper macro from `calibrate.cfg` and restoring the calibration macros to their original states ? I think it could be a good compromise.

d5aint:
Looking over the changes in 2.0 I have to agree. I see you are looking for more of a module approach.
I am not fully up to speed on modifying this with out closing it and starting over. Although it may be easier.
Right now it is way outside the scope of 2.0.

d5aint:
Reworking

PR #34, opened 2022-10-24
add extruder config files

fbeauKmi:
Here is a proposal of config files for different extruders:
/config/harware/base/extruder/base.cfg contains common parameters (Pins, hotend temperatures, pid, ...)
other files contains only specific datas for each extruder (gear_ratio, rotation_distance, run_current)
Did not deploy in printer cfg yet.

PR #37, opened 2022-10-25
Add extruders config files.

fbeauKmi:
Sorry for the previously closed PR, I did something wrong rebasing the PR #34 ;)
This is the same as previous + galileo extruder.
/config/harware/base/extruder/base.cfg contains common parameters (Pins, hotend temperatures, pid, ...)
other files contains only specific datas for each extruder (gear_ratio, rotation_distance, run_current)


Frix-x:
Hey thanks for doing this PR again and thanks for the additions of some more extruder config files.

Now before merging I would like to discuss with all the contributors of my config about the split in a "base file" + "specific file" : in my own opinion, this add some complexity for a potential user of this repo as there would be two file to include and to review when configuring the extruder. I think it's better to keep it the most simple possible with only one file per extruder configuration.
What do you think about that ?

Also, today we need to do that because Klipper does not support config variables for now. But what if we try to do a PR on Klipper to add this config variable support and try to make it merged ? I think it would be a better way to handle this problem.

What are your thought ?

Henlor:
This is a tough question in my opinion.

The most important thing to keep in mind is the ease of use for the end user (with little knowledge of coding). 

And this repo is an good answer to that use case in my opinion. Because it makes it easier to add or enable features and mods.
But not a perfect one, mostly because klipper is missing some features. 
The config variables to make the code simpler, cleaner and lighter.
And also better reporting for errors messages with at least the path of the file containing the error. To able to navigate and troubleshoot easily. 

And we need to be aware that each time we will add a new subdirectory, configuration file or splitted section declarations, this will certainly introduce more confusion for the end user during the first configuration phase and/or in case of error.

Honestly, I don't know which approach would be the best: Splitted section declararation or not. Or even, if we have to choose between this two options.

This approach shares the same indiscutable advantages in term of modularity as using multiple config files and subdirectory. But it also suffers from the same disadvantages by adding more complexity to the code.

Without testing it, I think this is too soon to be able to know precisely if this is a good or bad solution. 

And maybe the best solution would be to find the right balance between those different ways of modulate the code.

What's your thoughts?

fbeauKmi:
I don't have the right answer to your question. Actually, I splitted [extruder] for this PR as I do in my own config, I was probably a bit lazy. The difference is I splitted the config for hotends (HF, SF, ED3,...) as I added variables for nozzle, tip shaping, ...  but I'm not sure that it makes sense for now in your configuration. 

By the way, I like the way you made optional some features, making config portable on different printers and really powerfull, but I don't think end users without knowledge of Klipper should use it as they will have to check every hardware files that they do not understand. Discord helpdesk will explode XD

I did not read much on this part of Klipper's code but I imagine a PR on Klipper could be done effortless as it uses Configparser library (conjecture).

I follow your recommandations and remove base.cfg

cheers,

Frix-x (approved):
Ok for me for now.
So I'll merge it without the split but will keep it on my list for a potential future integration later if klipper variables are still not possible.

Thanks !

Frix-x:
I'm closing it now as it's merged in 2f433c0

Frix-x (commented):
I'm closing it now as it's merged in 2f433c0

PR #38, opened 2022-10-26
pinout map proposal

fbeauKmi:
Hi, 
here is a pinout map proposal according to the board_pins. I also removed duplicate pins (I don't know how I introduce that) and mistakes.
I should add a link to the board references (manufacturer documentation). PDF is more suitable for that
What is your thought?

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,13 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+lfx_crowdfunding: # Replace with a single LFX Crowdfunding project-name e.g., cloud-foundry
+custom: paypal.me/felixboisselier # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
--- a/config/variables.cfg
+++ b/config/variables.cfg
@@ -91,6 +91,7 @@ variable_disable_motors_in_end_print: True
 variable_purge_and_brush_enabled: True
 variable_brush_xyz: 77.5, 307, 2 # Position of the brush center for nozzle cleaning
 variable_purge_bucket_xyz: 20, 307, 5 # Purge bucket position
+variable_purge_distance: 30 # Amount to purge
 
 # Recirculating carbon filter (example: Nevermove v5, etc...)
 variable_filter_enabled: True
--- a/macros/calibration/pa_calibrate.cfg
+++ b/macros/calibration/pa_calibrate.cfg
@@ -104,9 +104,9 @@ gcode:
     {% set print_width    = num_raft_lines * spacing_raft %}
     # adjust print height to align with the band grid
     {% set band_height      = lines_per_band * line_width + (lines_per_band - 1) * spacing_line + spacing_band %}
-    {% set bands_per_height = (([print_size, max_y]|min  - spacing_purge) / band_height)|round(0, 'floor')|int %}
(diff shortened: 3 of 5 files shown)


Henlor:
Hey,
I had noticed there were some mismatched pins. But you are too quick ^^
It's also a nice touch to have a picture of the board. 

You can add this link for the manufacturer documentation.
https://github.com/bigtreetech/BIGTREETECH-OCTOPUS-V1.0/blob/2e0ab04eef0a3dc0e7caf15eb04169fd47efd1fd/Hardware/BIGTREETECH%20Octopus%20-%20PIN.pdf

And I think it could be interesting to add this link too.
https://docs.vorondesign.com/build/electrical/v2_octopus_wiring.html

There is only one thing that bothers me.
The pins of hotend_heater and bed_heater/SSR are reversed compared at the Voron documentation (link above)



And maybe add some variable configs for the probe in comment

`# Inductive Probe` 
`## Select the probe port by type:`
`PROBE_INPUT=PG15,     # DIAG7  - NEEDS BAT85 DIODE! Set pin value to "!PROBE_INPUT" in [probe] section if probe is NO.`
`#PROBE_INPUT=PB7,      # SENSOR - Will not work with Diode. May need pull-up resistor from signal to 24V. Set pin value to "~!PROBE_INPUT" in [probe] section.`
`#PROBE_INPUT=PC5,      # PROBE  - Probe port; NPN and PNP proximity switch types can be set by jumper. Set pin value to "~!PROBE_INPUT" in [probe] section. For Octopus Pro only.`

EricZimmerman:
here is a GIANT image of all the pins on octo board. i like it better than the PDF

https://github.com/EricZimmerman/VoronConfigs/blob/master/btt_octopus_1.1_pins.png

Henlor:
> here is a GIANT image of all the pins on octo board. i like it better than the PDF
> 
> https://github.com/EricZimmerman/VoronConfigs/blob/master/btt_octopus_1.1_pins.png

Personally, I prefer the PDF link to the BTT repo ^^

You are already on their repo if you need more informations.
And you can download the pdf if you need to zoom in it. 
And most importantly in a PDF, you can find a pin location by its name. This not possible in en a PNG file :)

EricZimmerman:
good points on searching. that is handy

Henlor:
This saves me from becoming nuts multiple times ^^'

Frix-x:
Hey thanks for this one !

Good idea to put some pinmaps in the repo. I agree that it's easier to ctrl+f in a pdf file to find the pins. Also git is able to render pdf directly online now. So IMHO best is to go with PDF files either on my repo or directly by a link to the manufacturer repo.

Regarding the pinmaps images on this repo, not sure about directly saying : "here you plug the heater", "here you plug xxx", "here you plug yyy". But I would prefer to just say : here is port "P1.24", ... As on the giant picture that @EricZimmerman linked above.
In fact my pins maps are the way I plug my machine and "an example" of how to do it. But the beauty of Klipper is that you can do whatever you want. So if you just put the ports name in the file, you let the users potentially change them easily and customize it how they want it to be. What do you think about it ?

fbeauKmi:
Hi,
@Henlor About bed pin choice,  I put it like this : first bed then hotend as it is done for thermistors. It differs from Voron doc, but this one often changes too ^^ . About probe pin : PG15 is common between Octopus and Octopus pro and work perfectly with mecanical probe. 
I'm agree whith a link to the PDF/SVG pinmap of official repo
@Frix-x I problably missed to explain. The aim of this pinmap was not to fix the pins usage. It's just to say to the end user " If you use the board_pins.cfg without to care about pins references  you MUST plug like this". Perhaps just add a warning :"This pin map is an example, if you plug another way refer to (link) and modify board_pins.cfg accordingly" or something like. 
I think too that the magic of klipper is to let tweak your mcu as you need !
A link to the boards official repo in the board pins file should be good option too. Let me know before I go further.

Frix-x:
Yes, so a link to the officials repos in the board_pins files is maybe the easiest solution as there is nothing else to do for us.

Then add some pictures as you have made is a good addition for the global documentation of the repo that I plan to start when back from my weekend.
FYI, I plan to make a "doc" folder at the root of the repo with a set of markdown files for everything (install instructions, functionnalities, etc...). Images could be added here as well. What do you think ?

fbeauKmi:
Hi @Frix-x , I'm quite busy for the next couple of weeks, I'll manage to do that soon. But if you're in a hurry do it XD

Best regards .

Frix-x:
No problem, I'm not in a hurry.
This is a nice addition but can be added later when the release of the config v2 is already done :)

Frix-x:
Hello, this PR was based on a branch that was just merged. Can you reopen it on main ?

PR #39, opened 2022-10-26
Refacto:  multiple mcu configuration management

Henlor:
This is a proposition on how to handle the pin_aliases for multiple mcu config.

WHAT THE GOAL OF THIS PR
As the most common case is to have a printer whit only one controller_board(mcu) and In order to make the most generic config as possible, the 'default' pin_aliases registered in the different files and sections should be used without prefix (case of multiple mcu). 

HOW IT WORKS.
All default pin-aliases are declared without prefix.
For the case of multiple mcu, a new file can be included in the printer.cfg. It contains all sections that contain a pin declaration with a prefix. As this file is called after all the previous 'default' pin déclarations. This should in theory  override them.

!!! This has not been tested as I don't have a multiple mcu config !!!

WHAT HAS BEEN MODIFIED
- remove all prefixed pin_aliases from config/
- create _multiple_mcu_pin_override.cfg file and 
- add an include in printer.cfg

- FIX: specific mcu pin called in some fan section replaced with their corresponding pin_aliases.

### DIFF
--- a/config/hardware/addons/accelerometers/adxl345.cfg
+++ b/config/hardware/addons/accelerometers/adxl345.cfg
@@ -1,6 +1,6 @@
 [adxl345]
 spi_bus: ssp1
-cs_pin: z:ADXL_CS
+cs_pin: ADXL_CS
 axes_map: -z,y,x
 
 [resonance_tester]
--- a/config/hardware/addons/filters/nevermore_filter.cfg
+++ b/config/hardware/addons/filters/nevermore_filter.cfg
@@ -9,7 +9,7 @@ gcode:
 
 
 [fan_generic filter]
-pin: z:FILTER_FAN
+pin: FILTER_FAN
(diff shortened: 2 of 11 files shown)


Frix-x:
Hey thanks for the idea and the proposal.
I've took some time to test it and I can confirm that it looks to be working. However it's not that easy for a beginner to do that and not plug and play as it involve to modify the overriden sections in the file t oadapt it to the machine... Not that easy !

So for the moment I'll keep this one on hold to see if there is a better way to do it. Indeed, I would prefer a klipper modification to be able to input directly the board name in the board_pins files or maybe use a correct config variable handling in klipper as an alternative (but this will be worked when the PR #28 is merged)

Frix-x:
Hello, this PR was based on a branch that was just merged. Can you reopen it on main ?

PR #48, opened 2022-11-22
add Pinout.md

fbeauKmi:
Hi, 
Here comes the pinout.md as it was told previously.
Are you okay with?


Frix-x:
Thanks for this PR ! You've done a very good job !

I'll just modifiy it a little bit the formatting (nothing bad, just to fit with all the rest of the docs) but I'm ok regarding all the content. Is it ok for you ?

PR #54, opened 2022-11-23
Create new ligth file for mini display 12864

zebons:
to define or change colors


zebons:

PR #55, opened 2022-11-23
add light file for mini display into printer.cfg

zebons:


zebons:

PR #56, opened 2022-11-25
Colors  management for Case & minidisplay leds and add error color

zebons:
Add action colors for mini display and case leds.
Add color for error (set into the on error gcode  commands)


zebons:

PR #60, opened 2022-11-30
Adaptive bed mesh v3

Frix-x:
This is a WIP about the v3 of the adaptive bed mesh !

Changes:
  - added the use of [exclude_object] tags to extract the first layer bounding box (many thanks to Kyleisah for the excellent idea and inspiration) the macro is still fully compatible with the old way using the SIZE parameter: it will use it if specified, or else fallback to the [exclude_object] method and if both are not available, it will do a full and normal bed mesh as usual.
  - also added a FORCE_MESH parameter to mesh even for very small parts
  - removed the RRI that was always added put in the BED_MESH_CALIBRATE call. Now it's added only when there is one defined in the [bed_mesh] section
  - changed the RESPOND for a better user experience with more understandable messages

Kayleisah what do you think about it ? Is it ok for you with the credits like that ? Thanks a lot for the idea of using the exclude_objects from Klipper !

### DIFF
--- a/doc/features/adaptive_bed_mesh.md
+++ b/doc/features/adaptive_bed_mesh.md
@@ -24,14 +24,14 @@ If you installed and use the full config folder of this github repository, this 
 If you want to install it to your own custom config, here is the way to go:
   1. Copy the [adaptive_bed_mesh.cfg](./../../macros/addons/adaptive_bed_mesh.cfg) macro file directly into your own config and include it.
   2. The macro needs the bounding box of the first layer to be able to compute a mesh on this surface. There is now two way to get it. Choose the one you prefer for your setup:
-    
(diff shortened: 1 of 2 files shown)


kyleisah:
Yeah this should work great! I’m glad it can help you!

Frix-x:
I just noticed something that I'll add in the documentation: using the [exclude_object] tags is a little bit less precise than using the original "SIZE" parameter directly from the slicer because of these two points:
- the exclude_object tags are using the full parts sizes (not only the first layer). So if you do a part with very large overhangs, it will do a big mesh with your method and will only mesh the base with mine. But this is marginal as it's still good to have a bigger mesh, just time consumming.
- if there is a skirt around the parts, it will not be took in account as there is no tags associated for this kind of "objects" and thus can lead to bad adhesion of the skirt in case of a very bad "taco" bed

So in conclusion, using [exclude_object] tag is better for new installations as it's much more easier to install. But for already working installation I would not advise to change and continue to use the SIZE parameter as the hard part is already done and will still work.
People can still update the macro to get the other benefits mentionned in the first post here :)

PR #68, opened 2022-12-06
CANBUS Support for Octopus & EBB36

W141-ID:
ebb36 board pins, compatible to the provided octopus can or other future configs


config/mcus/octopus.cfg
```diff
@@ -1,45 +0,0 @@
-[mcu]
-##--------------------------------------------------------------------
-serial: /dev/serial/by-id/change-me-to-the-correct-mcu-path
```
Frix-x, line 3: You deleted the original file in your commit. Can you bring it back please ?

Frix-x (changes requested):
Thanks for this one, it's a welcomed PR !

I think we can push it a little bit further to make it better.
Indeed, defining similar alias pins in multiple [board_pins] section is not a problem as at the end we use for example `toolhead:E_HEATER` instead of just `E_HEATER` and the physical port is still here.

So I would prefer if possible to keep the original full Octopus file (also valid for Spider boards) and add in the header : 
```
# This board work by using a serial connection by default.
# If you want to use CAN, invert the commented lines here

serial: /dev/serial/by-id/change-me-to-the-correct-mcu-path
# canbus_uuid: change-me-to-the-correct-canbus-id
```

Are you ok with this ?

config/mcus/octopus.cfg
W141-ID: Oh sorry this was by mistake. But because of shared PIN definitions between the CAN Board and the Octopus it's needed to have a second definition for the CAN config

W141-ID:
You see I reopened and closed the PR because I forgot that I changed a few config things please decide if it's okay for you or if I should just pull the two files

PR #70, opened 2022-12-06
CANBUS Support for Octopus & EBB36

W141-ID:
ebb36 board pins, compatible to the provided octopus can or other future configs
Sorry for the false PR's before. I am new to github.


W141-ID:
relocated led status to macro

PR #58, opened 2022-11-25
Manage Led Colors  Case & MiniDisplay

zebons:

PR #75, opened 2022-12-07
Tree simplification

Frix-x:
As stated in #58, this PR is there to simplify the folder structure of the config files


Frix-x:
Hey @zebons and @W141-ID, can you check if the folder structure is ok for you now ?

I just need to do some final testing tonight to check that no remaining old [include xxx] path are hiding in the config, but it should be good. I'll merge it afterward :)

zebons:
It's good for me. 
Mcus directory could be placed under hardware but no matter if you prefer to keep here.
Thanks to accept to simplify tree structure, it'll be easier to find files.

zebons:
I don't find the exhaust_fan.cfg file in the new tree structure

PR #77, opened 2022-12-09
CANbus Support (Ebb36 & Octopus)

W141-ID:


config/mcus/can_ebb36.cfg
```diff
@@ -0,0 +1,28 @@
+[mcu toolhead]
+# This board work by using a serial connection by default.
+# If you want to use CAN, invert the commented lines here
+
+# serial: /dev/serial/by-id/change-me-to-the-correct-mcu-path
+# canbus_uuid: change-me-to-the-correct-canbus-id
```
Frix-x, line 6: Can you replace the header ? This board is a dedicated CAN board and it will be more clean and similar to what we have in the other file `canboard_SHT36-42.cfg`.

```
##--------------------------------------------------------------------
canbus_uuid:  change-me-to-the-correct-can-id
##--------------------------------------------------------------------
```

BTW, can you also change the name of the file to `canboard_EBB36-42.cfg` ?

config/mcus/can_ebb36.cfg
```diff
@@ -0,0 +1,28 @@
...
+
+
+[board_pins mcu_EBBCan]
```
Frix-x, line 9: can you change the name here to `[board_pins mcu_ebb36_42]` ? This is to unify with the other canboard

config/mcus/can_octopus.cfg
```diff
@@ -0,0 +1,58 @@
...
+    EXP1_5=PE12, EXP1_6=PE13,    # Slot in the socket on this side
+    EXP1_7=PE14, EXP1_8=PE15,
+    EXP1_9=<GND>, EXP1_10=<5V>,
+
+# EXP2 header
+    EXP2_1=PA6, EXP2_2=PA5,
+    EXP2_3=PB1, EXP2_4=PA4,
+    EXP2_5=PB2, EXP2_6=PA7,      # Slot in the socket on this side
+    EXP2_7=PC15, EXP2_8=<RST>,
+    EXP2_9=<GND>, EXP2_10=<5V>
```
Frix-x, line 58: This file is not needed as someone could use the normal `octopus.cfg` file. So I think you can remove it :)

printer.cfg
```diff
@@ -11,6 +11,8 @@
 # [include config/mcus/octopus.cfg]
 
 # [include config/mcus/canboard_SHT36-42.cfg]
+# [include config/mcus/can_ebb36.cfg]
```
Frix-x, line 14: Please don't forget to change the name here to the correct new filename :)

Frix-x (changes requested):
Hey, thank your for this PR.
Now we can take some time to integrate it properly :)

config/mcus/can_octopus.cfg
W141-ID: Yes you are right, but since a lot of pins would move to the CANbus, won't it be confusing to the user?

Especially when changing pins to "toolhead:". Pins which are refered to the standard octopus won't pop up as fault and the user will have to search through different cfg files.

config/mcus/can_ebb36.cfg
W141-ID: should we destinguish between v1.1 and v1.0

Frix-x: Mhmm, I'm not familiar with these boards. Do you know if they have differences in the pinout ?

If it's the case, I would say yes, but if they are similar, I would say no.

config/mcus/can_octopus.cfg
W141-ID: Okay I have tought about it. Yes totally doable.
But you would need a documentation of every toolhead pinout documented.
Also you will need to make sure that the user understands which pinout are now free on the "mcu" board.

As you can see, because of the canbus. My fan & tempreature pins are different.

Just tell me which way you prefair and I will work on the documentation/comments.

It would also be possible to just comment which PINS should be commented/replaced when using Canbus.

Essentially combining both cfgs

Frix-x:
Hey. I saw that you just closed this PR.

I was quite busy yesterday but still ready to merge you additions if it's ok for you ? Can you reopen it ?

W141-ID:
Will do, had to change branches on my side.

config/mcus/canboard_octopus.cfg
```diff
@@ -0,0 +1,56 @@
...
+   EXP1_5=PE12, EXP1_6=PE13,    # Slot in the socket on this side
+   EXP1_7=PE14, EXP1_8=PE15,
+   EXP1_9=<GND>, EXP1_10=<5V>,
+
+# EXP2 header
+   EXP2_1=PA6, EXP2_2=PA5,
+   EXP2_3=PB1, EXP2_4=PA4,
+   EXP2_5=PB2, EXP2_6=PA7,      # Slot in the socket on this side
+   EXP2_7=PC15, EXP2_8=<RST>,
+   EXP2_9=<GND>, EXP2_10=<5V>
```
Frix-x, line 56: As stated in our last discussion: can you remove this file and just add the `canbus_uuid: change-me-to-the-correct-canbus-id` line to the original `octopus.cfg` file ?

Then I'll add some documentation on my side to the install note about it to say to add `toolhead:` for the pins of the extruder, hotend and everything that is plugged to the canboard if using one of the `canboard_xxx.cfg` file.

I think it's better to avoid duplicate to do it this way.

printer.cfg
```diff
@@ -10,8 +10,10 @@
 # [include config/mcus/spider_v2.cfg]
 # [include config/mcus/octopus.cfg]
 
-# [include config/mcus/rpi.cfg
+# [include config/mcus/rpi.cfg]
 # [include config/mcus/canboard_SHT36-42.cfg]
+# [include config/mcus/canboard_EBB36_42_v1.1.cfg]
+# [include config/mcus/canboard_octopus.cfg]
```
Frix-x, line 16: Please remove this line accordingly :)

Frix-x (changes requested):
Hey, so I think we are almost ready to merge it :)

Thanks for this PR !
Then we will be able to rework the LEDs as you wanted to do it

config/mcus/canboard_octopus.cfg
W141-ID: Alright, is it okay for you if I add those pins as comments in the octopus file? (Reference)

Frix-x: Yes I'm okay with that :)

Frix-x:
Oh yes, you're right ! I forgot about the exhaust_fan when doing the config refactoring !
Thanks for the addition ! Maybe we can find a free pin on each MCU to let it uncommented and available. This will avoid the config to crash if someone include the exhaust_fan.cfg file without uncomment the line in the board_pins file. The goal is to have the user have the less interaction with the inner config as much as possible.

Then I think I'm ready to merge it when you will be ready !

W141-ID:
@Frix-x I changed/added the pins. I have seen that all your configs don't make use of the bigger bed heater terminal. I wonder why? I changed it in the cfgs but can revert it at any time. also maybe someone that uses the fystec v1 board can add the pinout because I didn't had the time too look further into that board.

W141-ID:
> Hello, yes this is because on Voron printers we almost exclusively use an SSR to control the bed heating. This does not draw a lot of current and that's why we prefer to connect it to a fan port and let the big bed connector to something that need more current (like in my case for fcob leds, but it could be anything else)

Ah gota. Will revert it. I just use the bed terminal on my 2.4 out of convinience.

config/mcus/SKR1.4_x2_V2.4.cfg
```diff
@@ -74,6 +74,9 @@ aliases:
 # recirculating filter
     FILTER_FAN=P2.7,
 
+# EXHAUST FAN
+    EXHAUST_FAN=P2.4,
```
Frix-x, line 78: Can you change it to `P2.5` to avoid a duplicate pin (with z:CONTROLLER_FAN) ? Thanks :)

Frix-x (changes requested):
Just a little change, but I'm ok for it now :)

Frix-x (approved):

PR #78, opened 2022-12-09
Create a mcu rpi.cfg

zebons:


zebons:
since the rpi can be used by klipper as an auxiliary mcu for other functions than the adxl345 (sensors/leds connected on GPIO ports for example) it is preferable to create a specific mcu file for the rpi and to remove the mcu definition from the adxl345_rpi.cfg file

Frix-x:
Thanks ! I'm ok with this one as is :)

PR #86, opened 2022-12-13
fix relative extrusion potential bug in the end_print macro

zebons:
# Add an E92 E0 command to prevent error if the file printed was in relative extrusion mode

### DIFF
--- a/macros/base/end_print.cfg
+++ b/macros/base/end_print.cfg
@@ -17,6 +17,7 @@ gcode:
     {% else %}
         # pull back the filament a little bit
         _TIP_SHAPING
+		E92 E0				# Prevent error if the file printed was in relative extrusion mode
         G1 E-10 F2100
     {% endif %}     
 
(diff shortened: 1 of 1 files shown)


macros/base/end_print.cfg
```diff
@@ -17,6 +17,7 @@ gcode:
     {% else %}
         # pull back the filament a little bit
         _TIP_SHAPING
+		E92 E0				# Prevent error if the file printed was in relative extrusion mode
```
Frix-x, line 20: Can you replace the tabulation by spaces ? Thanks :)

Frix-x (changes requested):
small typo, but ok to merge it then :)

Frix-x (approved):

macros/base/end_print.cfg
```diff
@@ -17,6 +17,7 @@ gcode:
     {% else %}
         # pull back the filament a little bit
         _TIP_SHAPING
+        E92 E0
```
Frix-x, line 20: I think it's `G92 E0` that you want to use ?

PR #87, opened 2022-12-13
Rename of exhaust_fan.cfg to exhaust_filter.cfg

zebons:
Change location of exhaust_fan.cfg file from fans directory to filters


Frix-x:
Hey thanks for this one. I'm ready to merge it but I'll keep it on hold at the moment waiting for #77 to be merged. This is necessary to avoid a git conflict :)

Frix-x:
Hello,

#77 was just merged and I'm ready to work on this one.

You made some change regarding pins about filters and exhaust that I don't like that much:
  - You changed "filter" to "nevermore" and this break the modularity. I prefer to keep the generic "filter" term in case we add new filter hardware and let them work the same way as the current nevermore filter
  - You included the macros in the exhaust filter but this is not necessary as it's a temperature_fan. So this can be removed
  - You changed some pins and this was done partially in #77 and/or you changed the name to "NEVERMORE_FILTER" instead of the old generic name that was in place

So can you revert to [dea1786](https://github.com/Frix-x/klipper-voron-V2/pull/87/commits/dea17866bf27200605e26c711ee0945508396fe9) ? I'll merge it immediately and then we can open a new PR and discuss about the other changes if you want (but not sure they are a good idea at the moment)

Frix-x:
Merged in 29c1fe63340ce30607c622cdb6c256665869c128

PR #89, opened 2022-12-17
exhaust with or without filter and remind to refresh your carbon filter

Benoitone:
Some modifications to let use an exhaust fan with or without filter and adapt start and end_print for materials.
And include the MapleLeafMakers macros to remind to change carbon filter.


Benoitone:
Hello,
I tried to include the macros for the exhaust fan on V3 do you want me to PR or is it not worth it?

Frix-x:
Hello, I think this is still a good addition. I would be happy to accept your PR.

The only thing that I'm still not sure how to integrate it properly today is the time counter macros. Maybe finding a way to get it optional for an user would be better if possible :)

PR #101, opened 2023-01-08
Update probe_base.cfg

d5aint:
Removed the potential of attempting to probe the bed when the probe is not over the bed and would cause the toolhead to hit the bed.

### DIFF
--- a/config/variables.cfg
+++ b/config/variables.cfg
@@ -31,6 +31,10 @@ variable_accel_to_decel_factor: 0.75
 # Dockable probe variables
 #######################################
 
+# maximum Bed size avoids doing a probe_accuracy outside the bed
+variable_max_bed_y: 275
+variable_max_bed_x: 298
+
 # Minimum safe Z height to attach/detach probe
 variable_probe_min_z_travel: 15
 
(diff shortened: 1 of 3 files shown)


Frix-x:
Hello!

Thanks for this PR. It's a good idea :)
However two points:
  - If possible, can you add the probe offsets in the check as they will be used when the probe come back to the testing point ? 
  - I'm not very comfortable to hard set directly in the macro the +-8mm and +-45mm offsets to define the bed area. Maybe adding a variable in the variables.cfg that define the two opposite corners of the bed could do the trick but it's not really user friendly... Do you have another idea ?
  
As an example, here is a piece of code used in the current Klicky macros:
```
    # Protect against PROBE CALIBRATE performed from outside the bed
    {% if printer['gcode_move'].position.y > (max_y - probe_offset_y)
          or printer['gcode_move'].position.y < - probe_offset_y
          or printer['gcode_move'].position.x > (max_x - probe_offset_x)
          or printer['gcode_move'].position.x < - probe_offset_x %}
      { action_raise_error("Must perform PROBE_CALIBRATE with the probe above the BED, check klicky_variables bed size!") }
    {% endif %}
```
Here max_x and max_y are variables to define the max point of the bed (and they assume that the min point is `0, 0`)

d5aint:
Let me know what you think of these updates.

PR #62, opened 2022-12-02
DB Filament Management

zebons:
Add db_filament.cfg developped by fboc and mofified for integration into Frix macros. start_print.cfg file modified to add db filament management

PR #104, opened 2023-01-23
[WIP] v3.0.0 - Auto update & full genericity

Frix-x:
- Add automatic upgrade with moonraker update_manager
 - Improvement of the file structure to get a better static config structure on one side and all the user custom config overides on an other side. Now, every users will be able to modify their own files in their own config directory while keeping all files here as read-only and easily updatable
 - Voron-TAP support added
 - Optional G10/G11 calls (if fimware_retraction.cfg is included)
 - Optional Y axis extruder cleanup
 - Optional heaters off at end of print


elpopo-eng:
- add a clean nozzle without purge before bed_mesh -> a safe temp heating is necessary to have a good cleaning
- modify extruder heating sequence : here we can't load ercf at safe temp for bed probing
- change DOCK/ATTACH probe to something more generic : activate/disable so for tap we manage extruder temp in activate and reset to previous temp in disable
- for tap G28 Z is operate like z_calibration and maybe unaligned with relative_reference_index of bed_mesh

Frix-x:
I will also correct the print_start back to what it was before as I don't like to do the mesh before the rest : I prefer it at the very last moment as it was before.

yomgui1:
The heightmap disappear after the G28. Need to check why.

W141-ID:
Great additions!

For Tap the routine should always be performed hot (150C)

I tested out following setup and it worked out pretty good:

Home
Bed heating
Clean Nozzle, followed by nozzle cool-down to 150
Qgl
Full home again after qgl (important step!)
Bed Mesh
Heating nozzle
Print

I probably forgot something but it's just shematics

yomgui1:
> I will also correct the print_start back to what it was as I prefer to do the mesh at the very last moment as it was before.

With TAP you cannot do the bed mesh at the last moment as it was done before. The extruder temperature was too high.

Frix-x:
No problem with that.
In fact I have a plan to simplify this as there is lot of change now and I think all of this is not needed. So I plan to create new macros to get something like "prepare_probe" and "finish_probe". Thoose macros will work as an entry point each time a probing need to be done and will be in charge of:
  - deploying/retracting the dock using a servo if needed (in case it's a dockable probe with a servo docking system)
  - call the attach/dock the probe (in case it's a dockable probe)
  - deploy the servo (in case it's a BLTouch like probe)
  - change the temperature to the limit if it's currently above (for TAP mainly)
  - and restore the temperature at the end of a probing (if temperature was changed for TAP mainly)

This will avoid to get all the checks and everything that were added and simplify a lot as all the probing will be called the same way and will become more probe agnostic.

yomgui1:
> The heightmap disappear after the G28. Need to check why.

It looks that G28 Z erases the bed_mesh. So :
- I move the G28 Z before the bed_mesh
- or I save and restore the bed_mesh

Preferences? (my preference is the G28 before bed_mesh, I don't see advantages to make the Z home after a bed_mesh)

Frix-x:
I'm currently working on it right now. I'll push everything before tonight :)

Frix-x:
Ok, now I have set the print_start back to a good condition with the correct action order.

To proceed I added two macro: ACTIVATE_PROBE and DEACTIVATE_PROBE (with conditionnal include) to be able to set the printer in a correct state to probe (and restore it when probing is finish. ie attach probe, verify temperature, etc... depending of the devices installed). Thoose macros are also used for the G28 Z but only in the case of a TAP probe installed.

This commit also added the support of the standard inductive probes as well to be able to support stock printers 🎉

Frix-x:
Hey, @yomgui1 if you get to a working state, can you squash your last 6 commits (and force push) to try to keep something a little bit more clean now ?
Thanks :)

yomgui1:
> Hey, @yomgui1 if you get to a working state, can you squash your last 6 commits (and force push) to try to keep something a little bit more clean now ? Thanks :)

done

yomgui1:
So, I've warned on Discord. But preferable to talk about that here.
Current order of START_PRINT macro is not, for me, acceptable for TAP owners.
Keeping the V2 order as it with a TAP gives:

1. Homing (forced or not)
2. Bed keatsoak (at print level)
3. Chamber heatsoak (at print level)
4. QGL
5. Extruder heater (at print level), composed of:
   * M109 S{EXTRUDER_TEMP}
   * Optional ERCF setup
   * G28 Z  **=> generic probe macro causes here a M109 on a tap safe temperature, and when done, re-increase again at print temperature**
6. Noozle cleanup and _Z-Calibration_ **=> Z-calibration is not done with tap, only nozzle clean-up, but weird to found it here**
7. Material setup (pressure advance, retraction length and  gcode file Z-offset)
8. Bed-Mesh **=> generic probe macro causes here a M109 on a tap safe temperature, and when done, re-increase again at print temperature**
9. PRIME_LINE

As you can see, steps 5 and 8 cause a double extruder temperature back-and-forth.
When starting at a low temp and for ABS, with a "normal" extruder cartridge, it's 4 minutes wasted.

Possible order solution (almost as implemented by my genuine PR):
1. set extruder temp at tap safe temperature as soon as possible (first step, or at least before the QGL).
3. bed/chamber heatsoak
3. performs a minimal homing and do a nozzle clean-up at this safe temp before the QGL (some residues may corrupt the QGL)
4. QGL
5. Z-reset (G28 Z)
6. dynamic bed_mesh
7. Set extruder to print temp
8. Optional ERCF setup
9. Optional Z-calibration
10. Material Setup
11. PrimeLine

Some may think that the Z-reference was not made a print temp can issue a bad Z-calibration (step 9). But I think the dilatation offset between 150 and higher temp like 250 (or even more for tech filaments) is not significant and in last case may be (should be!) set by the gcode file (during the step 10).

Waiting for your comments.

zebons:
I take a look on your V3 and it's a great work for having a fully generic config.
However I notice an issue in your start_print macro. The material parameter part (section 7) isn't generic. So you cann't update automaticly that part. To resolve quickly that issue you can first add a material parameter part in your variables.cfg file like that :
# Material variable parameters

variable_pressure_advance: {
        'PLA': 0.0525,
        'ABS': 0.0480,
        'PET': 0.0650,
        'PC' : 0.0550,
        'XXX': 0.0500,
    }
variable_retract_length: {
        'PLA': 0.55,
        'ABS': 0.50,
        'PET': 1.40,
        'PC' : 0.35,
        'XXX': 0.50,
    }
variable_retract_speed: {
        'PLA': 40,
        'ABS': 40,
        'PET': 30,
        'PC' : 40,
        'XXX': 40,
    }
variable_unretract_extra_length: {
        'PLA': 0,
        'ABS': 0,
        'PET': 0,
        'PC' : 0.2,
        'XXX': 0,
    }
variable_unretract_speed: {
        'PLA': 30,
        'ABS': 30,
        'PET': 20,
        'PC' : 40,
        'XXX': 30,
    }

and change the section 7 in your start_print macro by :
    # 7 ----- MATERIAL PARAMETERS --------------------------------
    # Material dependant parameters like PA, firmware retraction, Z_offset, etc...
    RESPOND MSG="Material: {MATERIAL}"
	{% set PAD = printer["gcode_macro _USER_VARIABLES"].pressure_advance[MATERIAL]|default(0.05)|float %}
	{% set RTL = printer["gcode_macro _USER_VARIABLES"].retract_length[MATERIAL]|default(0.5)|float %}
	{% set RTS = printer["gcode_macro _USER_VARIABLES"].retract_speed[MATERIAL]|default(40)|int %}
	{% set UNL = printer["gcode_macro _USER_VARIABLES"].unretract_extra_length[MATERIAL]|default(0)|float %}
	{% set UNS = printer["gcode_macro _USER_VARIABLES"].unretract_speed[MATERIAL]|default(30)|int %}
	SET_PRESSURE_ADVANCE ADVANCE={PAD}
	SET_RETRACTION RETRACT_LENGTH={RTL} RETRACT_SPEED={RTS} UNRETRACT_EXTRA_LENGTH={UNL} UNRETRACT_SPEED={UNS}

yomgui1:
> I take a look on your V3 and it's a great work for having a fully generic config. However I notice an issue in your start_print macro. The material parameter part (section 7) isn't generic. So you cann't update automaticly that part. To resolve quickly that issue you can first add a material parameter part in your variables.cfg file like that 

True!
I've some advanced ideas for that. A sort of filaments DB to be clear.
But for a first try a simple (static) dictionary as you give, user customizable using the overrides feature., could be enough.

zebons:
There is already a db_filament management developped by another user that I proposed to integrate into that config in the v2 version (see PR #62) but no actions since. https://github.com/Frix-x/klipper-voron-V2/pull/62
That proposal in an interim (?) solution to have quickly a generic solution for the near v3 version.

Frix-x:
Ok, so here is my points on that. I'll work on it tonight.

> Keeping the V2 order as it with a TAP gives:
> 
> Homing (forced or not)
> Bed keatsoak (at print level)
> Chamber heatsoak (at print level)
> QGL
> Extruder heater (at print level), composed of:
> M109 S{EXTRUDER_TEMP}
> Optional ERCF setup
> G28 Z => generic probe macro causes here a M109 on a tap safe temperature, and when done, re-increase again at print temperature
> Noozle cleanup and Z-Calibration => Z-calibration is not done with tap, only nozzle clean-up, but weird to found it here
> Material setup (pressure advance, retraction length and gcode file Z-offset)
> Bed-Mesh => generic probe macro causes here a M109 on a tap safe temperature, and when done, re-increase again at print temperature
> PRIME_LINE

This order was indeed optimized for dockable probes and auto Z calibration and I understand that it can be a problem for TAP with the temperature changes.
However I still need to find a way as there is some imperatives:
  1. Do a nozzle clean **before** the Z calibration (only appliable for dockable probes) to get a clean nozzle. There is a similar need for the last G28 Z with TAP.
  2. The nozzle clean **need to be done at the print temperature** to do it efficiently (especially for specific materials like TPU)
  3. The ERCF setup **need to be done at print temperature** and probably after the material setup (need to check for this, maybe it's not that mandatory)
  4. The adaptive bed mesh need to be done **after** the auto Z calibration (for dockable probes). There is not so much needs when using TAP as there is no need for a specific RRI definition.
  5. The bed mesh should be done at the very last step of the print_start macro, just before running the print. The printer frame can take up to two hours to dilate and stabilize and it move a lot during the first half an hour of heating. This means that there could be high variations from when you do the mesh and when you use it then during the print if there is some time in between. This is especially true on enclosed machines without backers where we could even see Z offsets differences between the beginning and the end of the first layer when running full plates of parts (ie. a first layer of 10-15min). That's why I think it's very important to keep it at the very last moment, just before the print to limit this.

I have in my mind a fully configurable print_start with something like a variable to define the order of steps. But it would be quite heavy to implement. So I would prefer to avoid that if possible...
Something like:
`variable_print_start_order: "home", "heatsoak_bed", "tilting", "z_calib", "xxx", "yyy"`

I'll think about how to do it with minimal change in the order for actual users to avoid side problems and crashs. Maybe in a first step a check for probe_type on tap or dockable probe and do the corresponding steps accordingly... Not very efficient and beautiful but better to get everyone happy quickly before finding a real and propre solution.
@elpopo-eng I need your thoughs on this topic as you are also one of the major contributor here and using it on multiple machines, some with TAP and some with a dockable probe.

Regarding the material DB, I agree I can work on that also for the v3 release. But it's not on the priority list right now, let put that on the side for now and come back to it when all the other problems are solved :)

Frix-x:
> WARNING: current macro state causes the extruder to heat-up at print temp just after the G28 Z, but depending on your probing configuration, the Z can be very low (around few micrometers) and may causes bed surface marks.

Regarding this, I'll push a fix to add a Z hop if nozzle is to close to bed when restoring temperature in the DEACTIVATE_PROBE macro :)

yomgui1:
@Frix-x it looks that a TAP prode order is quite incompatible in some ways with a dockable probe order.
So yeah, finding a more modular (but KISS) solution is needed

macros/base/probing/generic_probe.cfg
```diff
@@ -0,0 +1,69 @@
...
+    
+    # If a Voron TAP probe is defined, then check the temperature and lower it if needed
+    {% elif probe_type_enabled == "vorontap" %}
+        {% set ACTUAL_TEMP = printer.extruder.temperature %}
+        {% set TARGET_TEMP = printer.extruder.target %}
+
+        SET_GCODE_VARIABLE MACRO=ACTIVATE_PROBE VARIABLE=temperature VALUE={TARGET_TEMP}
+
+        {% if TARGET_TEMP > tap_max_probing_temp %}
+            { action_respond_info('Extruder temperature target of %.1fC is too high for TAP probing, lowering to %.1fC' % (TARGET_TEMP, PROBE_TEMP)) }
```
alikaragoz, line 33: Hey guys, just passing by on this PR to point to a small error in this MACRO:
```suggestion
            { action_respond_info('Extruder temperature target of %.1fC is too high for TAP probing, lowering to %.1fC' % (TARGET_TEMP, tap_max_probing_temp)) }
```

Frix-x:
> @Frix-x it looks that a TAP prode order is quite incompatible in some ways with a dockable probe order. So yeah, finding a more modular (but KISS) solution is needed

Ok, I have some time to think and discuss this point with elpopo. I think that I now have a modular solution that should not cause any problem to a basic user (everything will just work), but that also allow an advanced user to customize the order if needed. I'll draft it at lunch today and see how it goes :)

zebons:
`Regarding the material DB, I agree I can work on that also for the v3 release. But it's not on the priority list right now, let put that on the side for now and come back to it when all the other problems are solved :)`

Ok it's not a priority, but do you implement an interim solution like that I proposed  here to have a generic and automaticly upgradable V3 config ?

Frix-x:
Yes, it's planned to do it **before** the release, but it will be one of the last step of the work :)

yomgui1:
@Frix-x For TAP, you've kept the z_offset module, causing a G28 Z command, *after* the extruder_heating sequence.
In case of possible issue during this command we may lead to a PEI heating at too hot temperature.
I'm not confident to keep it like that.
Has someone else with a TAP get the same feeling?

Frix-x:
> @Frix-x For TAP, you've kept the z_offset module, causing a G28 Z command, **after** the extruder_heating sequence. In case of possible issue during this command we may lead to a PEI heating at too hot temperature. I'm not confident to keep it like that. Has someone else with a TAP get the same feeling?

Oops, my mistake. I forgot about this one and will move it before the extruder_heating if you want.

But how are you sure your nozzle is clean before doing the G28 Z then ? Using only the "safe temperature" cannot work for specific filament like TPU or PETG for example so that's why I think we have to let it like that as it's a safe default for a correct Z offset measurement.

Also keep in mind that there is a safety in the TAP activation procedure were it will lower the temperature to the safe temp before probing anything. So there is no risk for the PEI here, just some time to wait for the temperature to come to 150C before probing, then back to normal temp.

And at the end, if some people don't like that, because of the now modular approach, they can override it and define another order by themselves as they want it to be.

I will maybe need to split the "purge_clean" module in a "purge" and a "clean" module to allow you just a clean without having to also purge (and then use the material temperature). What do you think about that ?

W141-ID:
> I will maybe need to split the "purge_clean" module in a "purge" and a "clean" module to allow you just a clean without having to also purge (and then use the material temperature). What do you think about that ?

Splitting is a good idea! One could do a first nozzle cleaning before taping and after all the tapping a second cleaning + purging at extruder temp and then right into printing.

Frix-x:
Ok, now I've corrected the bugs of the modular START_PRINT, everything should work.

I've also splitted the purge and clean in two modules that can be called separately.
I also changed the TAP sequence order to put a clean before the QGL and Z offset, etc... and kept the purge and clean and the end :)
Finally I've set the G28 Z before a brush to be optional from a variable

yomgui1:
> But how are you sure your nozzle is clean before doing the G28 Z then ? Using only the "safe temperature" cannot work for specific filament like TPU or PETG for example so that's why I think we have to let it like that as it's a safe default for a correct Z offset measurement.

This is a permanent issue every where: even at the beginning we do a homing and a QGL on a Voron 2.4 for example, but even at this stage we may encounter some debris/filament oozing of a previous print, causing wrong QGL.
That's a tricky part.

yomgui1:
> I will maybe need to split the "purge_clean" module in a "purge" and a "clean" module to allow you just a clean without having to also purge (and then use the material temperature). What do you think about that ?

I +1 this idea!

config/mcus/octopus.cfg
```diff
@@ -1,59 +1,65 @@
...
+    EXP1_7=PE14, EXP1_8=PE15,
+    EXP1_9=<GND>, EXP1_10=<5V>,
+
+    # EXP2 header
+    EXP2_1=PA6, EXP2_2=PA5,
+    EXP2_3=PB1, EXP2_4=PA4,
+    EXP2_5=PB2, EXP2_6=PA7,      # Slot in the socket on this side
+    EXP2_7=PC15, EXP2_8=<RST>,
+    EXP2_9=<GND>, EXP2_10=<5V>,
+
```
Frix-x, line 65: I think you made a mistake on this file as everything looks to be changed even if just a couple of things were added. Maybe the line endings changed ? Can you check and correct it ?

yomgui1: it's not a mistake, all EOL on this file were DOS, I've converted it to UNIX (and added more pins)

Frix-x: Ohhh, this make sense. Thanks for converting it to proper lines ending then :)

Frix-x:
Hello all, I've pushed a couple of changes in order to get a clean MCU system. Please have a look at this (already merged here) PR: #120 

This allow a pretty cool mechanism that work in two stages:
  1. The user include the "standard" board_pins in his `printer.cfg` file. This file make the link between my "simple" pins names (ie. `PART_FAN`, `CHAMBER_TEMPERATURE`, ...) used everywhere in the config files, with the pin names printed on the MCUs PCBs (ie. `MOTOR0`, `HE0`, ...). This step also include automatically at the same time a "manufacturer" board_pins file.
  2. The second stage is done by the "manufacturer" board_pins file that do the link between what is printed on the PCB and the real chip pins (ie `P2.4`, `PB13`, ...) that is hard to handle for a standard user.

Also, the first stage files included by the user are configured to follow the "standard" way of wiring a printer as best as I could. But if for some reasons, someone is not happy and wants to wire his machine differently, then he could just add some overrides in his mcu.cfg file with entries like `PART_FAN=FAN5` instead of `FAN2` for example.

danielfmo:
I'm trying this now that I have multiple printers (never used your previous versions). 
I'm struggling to find a way to use this config with stepper drivers that are not tmc2209. 
Any hints?

Frix-x:
Hi @danielfmo ! At the moment the v3 branch is not ready for production and you should be ready for some bugs ^^

I currently have not implemented any mechanism to manage other stepper drivers than the standard TMC2209. So if you wants to use something else by keeping the "generic and read-only way" of the v3 branch, I have no solution at the moment other than adding the file by yourself.

But if you do so, I'll be happy to accept a PR and integrate this functionality :)

danielfmo:
@Frix-x completly understand that and that is my goal. 
I wanted just to ask if you had already thought in an implementation or  have any suggestions for me to implement it as at the moment I struggle to find an elegant way to implement it due to default_wiring.cfg already containing the tmc2209 section.

Frix-x:
Hello all,
I'm very happy with the last commit here. It add a new part in the install script that is called when it detect a new installation of the config (ie. no .VERSION file present in the user config directory) to:
  1. Ask the user if he wants to install some default MCU templates (or if he wants to do it manually)
  2. If answered yes, ask the user for the main MCU he is using and install the corresponding [board_pins] in his mcu.cfg file
  3. Ask the user if he is using a toolhead MCU (like a CANbus toolhead board) and install the corresponding [board_pins] alongside some default overrides to make it almost plug'n'play (in mcu.cfg and overrides.cfg respectively)
  4. Ask the user if he is using an ERCF MCU and install the corresponding [board_pins] if needed

In this commit there is also a lot of change in the comments of all the files to make it more clean and almost ready for the full release. I think we still need some general documentation in all the README files but now we are almost ready :)

zebons:
Hi,
I've had a look at the filament management integration and I think it's a good basis for having a modular management of filament settings. I was using this method before I tried to integrate with the fboc developer, but as the fboc developer is not very keen on this integration, this would already be a good basis for modularity. 
However, I have a few improvements to make:
First I don't see the default value for the startprint_actions variable (It is described in the _user_variables.cfg file as a comment but no default value...)
The filament parameters should be loaded at the beginning of the start_print macro because some parameters (soak times for example) should be different depending on the filament.
In the same way, it should be possible to set the temperatures (bed, extruder and chamber for example) at the level of the start_print macro. 
Why not add other filament-specific parameters such as the flow parameter. The parameters identified by fboc and previously integrated into PR#62 would be a good starting point. 
Of course these parameters could be managed at the start_gcode level of the slicer, but the objective of this centralised management should be to bring this management back to the level of klipper (and why not imagine using the same gcode file to print an object in PLA or ABS by simply changing the MATERIAL variable in the STRAT_PRINT line of the gcode file... let's be crazy).

yomgui1:
> The filament parameters should be loaded at the beginning of the start_print macro because some parameters (soak times for example) should be different depending on the filament.

Good point.

> Why not add other filament-specific parameters such as the flow parameter.

My point of view: the major reworks done by patch is not around the material DB. There are a lot of (more major) changes that need to be handled by users first. The current implementation is "enough" to have a good working start point. And one of V3 features is to make easier future evolutions of such things.
So I 100% agree with our improvement but I think it could be done later to let others features to be «accepted» by the community first.

Frix-x:
> My point of view: the major reworks done by patch is not around the material DB. There are a lot of (more major) changes that need to be handled by users first. The current implementation is "enough" to have a good working start point. And one of V3 features is to make easier future evolutions of such things.
> So I 100% agree with our improvement but I think it could be done later to let others features to be «accepted» by the community first.

This is exactly my point. I just want to get something that just work using the same "base" and config as the v2 branch while being able to customize and let the repo read-only (v3 system).

However, I already think about that and the way I just designed the material parameters enable it to be extended later to whatever we need. We have to keep in mind that this config is designed to be used by everyone, and especially novice users. So I would prefer in a first step to keep everything working as they already know it on classic START_PRINT configurations (ie. using flow and temperatures from slicer) and see how it can evolve in the future. A good thing would be to be able to activate/deactivate it from the variable.cfg file for example.

> The filament parameters should be loaded at the beginning of the start_print macro because some parameters (soak times for example) should be different depending on the filament.

Agree with that, but as said just before, I prefer to keep that for later to avoid changing everything here.

> First I don't see the default value for the startprint_actions variable (It is described in the _user_variables.cfg file as a comment but no default value...)

This is directly in the probe include files as the defaults are different depending of the probe configuration.

zebons:
> This is directly in the probe include files as the defaults are different depending of the probe configuration.

Ok, I hadn't considered that, which is perfectly logical since the order of the actions depends directly on the probe used

> >The filament parameters should be loaded at the beginning of the start_print macro because some parameters (soak times for example) should be different depending on the filament.

> Agree with that, but as said just before, I prefer to keep that for later to avoid changing everything here.

I agree with you, the proposed improvements can be added later. However, the filament parameters should already be loaded at the beginning of the start_print for 2 major reasons:
- We need to differentiate the loading of the parameters which must be done at the beginning of the macro for all the parameters (as is already done for most of them) from the use of these parameters which will be done later through the action. In my opinion the loading of the filament parameters should be done at the beginning of the start_print with the other parameters 
- This will avoid disturbing v3 users when later on they will be explained that this parameter loading must be done again because it could be used at different stages of the actions. It will also greatly simplify future additions of new parameters that could impact the different actions because they are already loaded at the beginning of the macro.

yomgui1:
> However, the filament parameters should already be loaded at the beginning of the start_print for 2 major reasons

Adding another good reasons: when you have not added a new filament, you can fail fast without all the calibration process (exactly what I've get right now...)

Frix-x:
The "fail fast" argument is the one that convince me to do this load at the beggining. But for the rest, we will let it like that for now for the v3.0.0.

I'll do the change :)

zebons:
Thanks and don't worry for 'the rest'. 
If the filament parameters are loaded at the begining, it'd be easier for us to do some enhacements in our config (for example adapt heatsoak time per filament) and propose to you new  PRs  for the future

Frix-x:
Ok, it's done. But the mechanism is the same.
I just removed the material module from the START_PRINT actions as this has no sense as it will need to be done at the beginning for some parameters and at the end just before the print for the others. Si I decided to remove the module and split the things in the START_PRINT directly

zebons:
it's ok for me for the release of v3.0 but I think that we may have to add the filament action for some parameters to position them only at the end of the macro (while keeping their loading at the beginning). For example if we decide later to store the hotend temperature per filament in the parameters, setting it at the beginning might be a problem for tap users ?

Frix-x:
Ok good news. Now I think we are ready for a full review before the merge. Can everybody here can have a look at the last version and see if everything is still working for him ?
That way I hope that we will be able to cover a good amount of configuration and check that it works for most of us :)

PR #106, opened 2023-01-24
QGL points and  gantry Corners per default config

barecool:
Suggestion for qgl points and gantry corner to be moved to each respective default config (per printer size / type) 
setup per standard voron config

### DIFF
--- a/config/hardware/default_250mm.cfg
+++ b/config/hardware/default_250mm.cfg
@@ -13,3 +13,12 @@ position_endstop: 0
 position_max: 210
 position_min: -5
 
+gantry_corners:
+   -60,-10
+   310, 320
+#  Probe points
+points:
+   50,25
+   50,175
+   200,175
+   200,25
--- a/config/hardware/default_300mm.cfg
+++ b/config/hardware/default_300mm.cfg
@@ -13,3 +13,12 @@ position_endstop: 0
 position_max: 260
 position_min: -5
 
+gantry_corners:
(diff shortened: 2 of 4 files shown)


Frix-x:
Hello,

Thanks for the suggestion, however this cannot work as there is not only QGL machines. Some user of this config (including myself) are also using Z_TILT (for Voron Trident or TriZero), or just nothing (for a Voron V0 for example). Also there is all the CoreXY machines from other brands which have other corners positions (I have at least one user using this config on a Vzbot printer).

barecool:
Thank you for taking look at my request - I totally understand 

I wish klipper would add the PR below - it would make your config 100 times more modular 
[klippy.configfile: Add conditional include #5003](https://github.com/Klipper3d/klipper/pull/5003)

Frix-x:
This and/or even being able to use Jinja directly in the config parser would be such a nice addition!
But I think Kevin is not really open on this kind of modification as he don't see the use-case. Maybe showing him my repo when v3.0.0 is out and telling him that 300+ people are currently using it right now (maybe more later) could change his mind?

Let's see!

Frix-x:
As I've not found a correct way to do it. I've added a "dirty" way in commit https://github.com/Frix-x/klipper-voron-V2/pull/104/commits/ad19c0cb737fb460d3941dfee7d9a14b5e9f8822

This will work until Klipper allow conditionnal includes or another system to do it easily :)

PR #108, opened 2023-01-26
V3 For Tap

Benoitone:
problems with Tap with 
endstop_pin , position_endstop and probe pin


Frix-x:
Thanks for this PR.

Your help on this V3 is very much appreciated.
As you've already seen (and corrected for the GalileoZ), we were focused on the "standard" hardware config for now but we need to convert all the other options to this "pin overide mechanism" such as all the XY, Z and extruder hardware files.

This is already a planned work to do before the release of this branch, but if you want to do it for the other files, I'll be happy to accept other PRs :)

Thanks!

PR #109, opened 2023-01-26
V3 XY Extruders and ercf board creation

Benoitone:


config/hardware/extruder/lgx_lite.cfg
```diff
@@ -9,6 +9,7 @@ full_steps_per_rotation: 200
 nozzle_diameter: 0.400
 filament_diameter: 1.75
 max_extrude_only_distance: 110
+sensor_type: ATC Semitec 104GT-2
```
yomgui1, line 12: Just for the note, temperature sensor is not something related to the lgx lite. But for sure klipper mixes that into the same extruder section. I would prefer to move that elsewhere (it's a pure user definition, as none has the same one). But as we need to do that everywhere, I'm fine with that for now

yomgui1 (commented):
Review by eyes is ok. But I need to test it first (not for ERCF parts as mine waiting for it's electronic :-) ). I'll do soon after my current print.

Benoitone:
For the ercf, I think that moggieuk's work is currently the most successful and perhaps trying to integrate it is a better solution than the current one?

config/hardware/extruder/lgx_lite.cfg
Benoitone: Yes, maybe gather all the sensor type in the same file maybe a solution, but given the great variability of the configs for this type of component it will most certainly end up in the overrides...

Frix-x:
Hey,

Thanks for the review. I'll merge it now as is and we'll work on the temperature sensor change when we'll do the final cleanup.

Regarding the ERCF macros, I'm not against a change, but this will need to be discussed in another PR I think :)

Frix-x (approved):

PR #118, opened 2023-02-17
Update temp_check.cfg

d5aint:
If temp is set at 150, then the variable T will never be set. The added elseif adds a check to see if the current temp is less than the one in T variable. This can/should be added to the main branch as well.

### DIFF
--- a/macros/helpers/temp_check.cfg
+++ b/macros/helpers/temp_check.cfg
@@ -5,7 +5,9 @@ gcode:
 
     {% if printer.extruder.target != 0 %}
         {% if printer.extruder.temperature < printer.extruder.target %}
-            M109 S{printer.extruder.target|float} 
+            M109 S{printer.extruder.target|float}
+        {% elif printer.extruder.target < T %}
+            M109 S{T}
         {% endif %}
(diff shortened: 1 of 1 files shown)


Frix-x:
Hey thanks for the suggestion!

Well, this change is not really how I designed the macro and use it today.
Indeed, in my macros I only use it as a check of the nozzle temperature:
  - If the nozzle is cold, then I set it to T temperature
  - If the nozzle is already hot, nothing need to be done as I don't want to change the already set temperature (this is to avoid messing with the specific temperatures needed in the other macros calling this _LOW_TEMP_CHECK).

Is it something not working for you ?

d5aint:
If I have the extruder set to 150, then do a manual purge calling [gcode_macro PURGE] giving TEMP=260 for the temperature, in [gcode_macro _LOW_TEMP_CHECK] the T variable will be disregarded since the extruder is currently set for 150 and the current temperature is 150.

Another approach would be to set the extruder temp in [gcode_macro PURGE] before calling the  [gcode_macro _LOW_TEMP_CHECK] macro.

PR #121, opened 2023-02-22
V3.0.0 MCU Fysetc and Mellow ercf boards and SKR1.4 modifications for ERCF.

Benoitone:
2 ercf boards
Rename runout_sensor of toolhead boards to toolhead_sensor
Mod double SKR 1.4 for ERCF


config/mcus/manufacturer/Fysetc_ERCF_ERB.cfg
```diff
@@ -0,0 +1,24 @@
+[board_pins mcu_ercf]
```
Frix-x, line 1: Can you change the name of the MCU in the manufacturer file to something else ? I'm not sure it will work as is as this could result in a conflict if for some reason two pin share the same name between my own pin notation and what is written on the PCB (ie. something like SERVO=SERVO would cause trouble).

So I would prefer to change it here to something like `ercf_manufacturer` or whatever you think is better.

config/mcus/manufacturer/Mellow_fly_ERCF.cfg
```diff
@@ -0,0 +1,25 @@
+[board_pins mcu_ercf]
```
Frix-x, line 1: Can you change the name of the MCU in the manufacturer file to something else ? I'm not sure it will work as is as this could result in a conflict if for some reason two pin share the same name between my own pin notation and what is written on the PCB (ie. something like SERVO=SERVO would cause trouble).

So I would prefer to change it here to something like `ercf_manufacturer` or whatever you think is better.

Frix-x (changes requested):
Thanks for the PR.
I'll merge it when the change is done about the `[board_pins name]` as there is currently a risk of conflicts if we keep the same name between the manufacturer and the user boards pins files.

Frix-x (approved):

Frix-x:
Thanks, I'll merge :)

PR #123, opened 2023-02-27
Update hotend_fan.cfg

Benoitone:
what about start hotend fan when emergency stop occur? To prevent deformation of the toolhead


Frix-x:
This is already the case : by default (if ommitted), a `[heater_fan]` has a `shutdown_speed` equal to its `max_power`

PR #128, opened 2023-03-16
fix BELT_SHAPER_CALIBRATION fail to generate png graph due to empty csv files

ndanyluk:
The belt_shaper_calibration and axes_shaper_calibration macros are implemented such that the underlying klipper macros write raw data to /tmp.  However, klipper generally creates a file in /tmp before writing data to it.  This can result in a race condition, where plot_graphs.sh can find the file and move it out of /tmp before klipper has finished writing data to the file.

This commit resolves this race condition by implementing a check to see if the file is open using the linux "lsof" utility.  Although I've only seen this issue on belts_shaper_calibrate, I expect the underlying problem exists on all 3 macros.

This would fix bug: https://github.com/Frix-x/klipper-voron-V2/issues/107

### DIFF
--- a/scripts/plot_graphs.sh
+++ b/scripts/plot_graphs.sh
@@ -6,6 +6,7 @@
 # @version: 1.5
 
 # CHANGELOG:
+#   v1.6: fixed race condition that results from input shaper macros delayed writes to .csv raw data in /tmp.  Added dependency for "lsof" utility.
 #   v1.5: fixed klipper unnexpected fail at the end of the execution, even if graphs were correctly generated (unicode decode error fixed)
 #   v1.4: added the ~/klipper dir parameter to the call of graph_vibrations.py for a better user handling (in case user is not "pi")
 #   v1.3: some documentation improvement regarding the line endings that needs to be LF for this file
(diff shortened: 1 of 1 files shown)


ndanyluk:
One thing to note is that "lsof" is not normally installed in Debian, so this does add a dependency that wasn't required previously.

Frix-x:
Hello, I wanted to try another idea without the use of lsof. Can you test the last version of the script that is pushed in https://github.com/Frix-x/klippain/commit/f1cf77a1f3381e749403b3c0a9b279db2d5db402 ?

FYI, I just added a sync && sleep 5 before the use of any function. I hope it will be sufficient to solve the problem

Frix-x:
Ok, the dirty fix was apparently not working, so I tried another way: simulating the bahavior of lsof, but without lsof.
Can you trey it and let me know if it works for you? See this commit: https://github.com/Frix-x/klippain/commit/e40edf2a26167bbbac07652c28c9ca312c61c51e

Frix-x:
I got two confirmed case that https://github.com/Frix-x/klippain/commit/e40edf2a26167bbbac07652c28c9ca312c61c51e has fixed the problem. So I'll close this PR now.

Thank you for the diagnosis and the idea of waiting for the file to be closed by Klipper ! It helped me a lot :)

PR #129, opened 2023-03-19
Input Shaper documentation

Frix-x:
This is a WIP of some documentation about how to read the IS graphs


Frix-x:
@ironhalik and @SinisterRj you wanted to be part of that. Here is a draft PR about some documentation.

At the moment, I've done all the theory explanation about how input shaping is working and on what parameters it is acting. I also added explanation about belt graphs but I still need to work on the standard input shaper graphs now and also vibration graphs. In the meantime, if you have some comments on what is already written, just tell me.

PS: don't look at the comments in the file as they are my "draft and source on which I need to work" :)

PR #130, opened 2023-03-25
Sensorless homing improvements

Frix-x:


AMFkuna:
Hi,
shouldn't it be homing_retract_dist: 0 in sensorless.cfg? Should be homing_retract_dist: 5 when using endstops.

Frix-x:
No, in fact it was already not used when using standard endstops as I use a mechanism in the homing override to use variables. This allow me to also avoid the dock or place the toolhead directly on the Z endstop pin if needed in only one move command.

See these variables:
https://github.com/Frix-x/klipper-voron-V2/blob/38a9279bff91a0acc7e2f6f793dd3b2e8b300b68/user_templates/variables.cfg#L44-L46

So it was set to a small value before to have a "minimum" backoff in case the user put 0 in the variables, but at the end it's not really a problem in fact so this safety can be removed and it make it also more easy at the same time for sensorless homing :)

PR #133, opened 2023-03-29
Extract basename before applying RegEx

spikeygg:
Resolves https://github.com/Frix-x/klipper-voron-V2/issues/131


Frix-x:
I'm sorry, due to the release of the new version, you will need to fork again the repo as I changed its name. Sorry for that :/

PR #136, opened 2023-04-06
Updated install.sh

W141-ID:
As I am using one PC for multiple Klipper instances, it was "difficult" to install the config without having a few problems.

By adding the function, the user can choose a custom folder name.
He can always decline if he has only one instance and needs to go the standard route.

I used the following scheme for naming as kiauh does the same:
Services as moonraker & klipper: klipper-xyz
config-data: xyz_data

I can change the function to ask to put in the full folder name.
I hope this addition is helpful to you.

I have tested it on my machine and it worked so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Installing Klippain should not be too complicated if you are already familiar wi
 
 Then, to run the installation script, connect to your printer using SSH and type the following command:
 ```
-wget -O - https://raw.githubusercontent.com/Frix-x/klippain/main/install.sh | bash
+wget -O - https://raw.githubusercontent.com/W141-ID/klippain/main/install.sh | bash
 ```
   
 This script will backup your old configuration, download this GitHub repository into your RaspberryPi home directory and setup the entire environment in `~/printer_data/config`. You'll also be asked if you want to select and install some MCU board_pins templates. This is recommended as it will allow a very fast filling of your `mcu.cfg` user file, but you can always do it manually afterwards if you prefer.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Installing Klippain should not be too complicated if you are already familiar wi
 
 Then, to run the installation script, connect to your printer using SSH and type the following command:
 ```
-wget -O - https://raw.githubusercontent.com/W141-ID/klippain/main/install.sh | bash
+wget -O - https://raw.githubusercontent.com/Frix-x/klippain/main/install.sh | bash
 ```
   
 This script will backup your old configuration, download this GitHub repository into your RaspberryPi home directory and setup the entire environment in `~/printer_data/config`. You'll also be asked if you want to select and install some MCU board_pins templates. This is recommended as it will allow a very fast filling of your `mcu.cfg` user file, but you can always do it manually afterwards if you prefer.
--- a/install.sh
+++ b/install.sh
@@ -25,21 +25,6 @@ BACKUP_PATH="${HOME}/klippain_config_backups"
 set -eu
 export LC_ALL=C
 
-# Step 1: Verify that the script is not run as root and Klipper is installed
-function preflight_checks {
-    if [ "$EUID" -eq 0 ]; then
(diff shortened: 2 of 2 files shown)


Frix-x:
Thanks you very much for this PR! I just discussed with someone about this :)

However I think it needs a little bit more work as I think it break the backup feature of the script. Don't forget this script is called everytime an update is done by moonraker. Do you have tested this part ?

W141-ID:
So creating a backup works actually fine. But should also be tested by someone else.
I will wait for the next update and check it out again.

I will have to check the updating part the next time an update is available.

But I think you are referring to the fact that anytime the install.sh script is called, there is no custom folder name available and it will stop or got the standard route right?

Frix-x:
> So creating a backup works actually fine. But should also be tested by someone else. I will wait for the next update and check it out again.
> 
> I will have to check the updating part the next time an update is available.
> 
> But I think you are referring to the fact that anytime the install.sh script is called, there is no custom folder name available and it will stop or got the standard route right?

Yes, in fact I'm referring to how it will handle the multiple printer thing: when you click the update button on one of the printers WebUi, moonraker will update the repo (doing a `git pull`), then will call the install script for this specific printer instance. However there is some things coming to my mind:
  1. Fluidd/Mainsail will not allow you to answer questions at this time (so everything should be automatic or will fail)
  2. The repo will be updated and so are all the symlinks on all the instances. So you will have all the printer updated at once, but the backup on only one of the instances

W141-ID:
This is most definitely true. It needs to be implemented into the install script or another way.

Another way is to have the install script get parameters from a local file?
It's only two variables that need to be "imported".

By importing those the install.sh can stay the same and be updated however you/we want to.

Frix-x:
> Another way is to have the install script get parameters from a local file? It's only two variables that need to be "imported".

This is indeed what I currently have in mind. You could add a `.CONFIG` entry to the .gitignore to avoid tracking it when created directly in the the Klippain repository path.

Then in the `install.sh` script use `~/printer_data/current/config/path >> .CONFIG` every time it's run to install a new printer.
Finally, when updating from one of the WebUI, make the script read this file and backup all the folders from all the machines one by one.

I think this is a way to make it work properly. What do you think about it? :)

W141-ID:
> > Another way is to have the install script get parameters from a local file? It's only two variables that need to be "imported".
> 
> This is indeed what I currently have in mind. You could add a `.CONFIG` entry to the .gitignore to avoid tracking it when created directly in the the Klippain repository path.
> 
> Then in the `install.sh` script use `~/printer_data/current/config/path >> .CONFIG` every time it's run to install a new printer. Finally, when updating from one of the WebUI, make the script read this file and backup all the folders from all the machines one by one.
> 
> I think this is a way to make it work properly. What do you think about it? :)

Have to read into it. But seems like a proper solution to me!
The ` .CONFIG` could be also used for future purposes.

Schematics:
1.  `install.sh` looks if there is a `.CONFIG` file

- if not it wil ask you if you need a custom path or just the nomal way

- if yes it get the custom path way and just do an update in that case

The `.CONFIG` should be created either way, custom or non-custom.

W141-ID:
Yeah I have seen that bug on the first install of one of my machines. It can leave quite a mess behind when updating, since adding/changing new variables will make the cfg no to work.

Not related to this pull, but be careful changing z-offset settings in the releases. At some point z-offset was changed from commented to uncommented. Messed up the z height and drove the nozzle throug the flex plate. Just a note to give other useres a warning when updating.

install.sh and update.sh would be even better. Integrating it into the kiauh worklfow could open the possibelity to grap machine locations automaticly and update/maintaining everything through kiauh.

Surion79:
what about putting the z-offset in the printer.cfg so it can be updated by klipper and it is sure that it is even correct after upgrade since it is in a `#*#` at the end of the config?

mjonuschat:
I'm closing this stale PR since there hasn't been any activity in a long time. The discussion indicates that more design of the feature is required, if needed we should start with a feature request to hash out the details.

Issue #137, opened 2023-04-06
Sensorless homing problem


kurtblah:
I get this error msg in the console upon start-up:
Error evaluating 'homing_override:gcode': jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'sensorless_current_factor'

This happened after moving to klippain from your earlier version. Not sure where to look, the klippy log looks normal to me.

My machine is stock 2.4, with TAP and Stealthburner. normal endstops installed.

Frix-x:
Hi,

Thanks for opening this issue!
Yes this is due to a new variable added in the last version: `sensorless_current_factor`

You need to add it in your variables.cfg file and everything will be alright (see the variables.cfg templates for the definition and default value)

kurtblah:
thanks, that solved it.

Frix-x:
> adding to that seems like the sensorless homing variables need to be floats. I use sensorless homing on my machine and had to change those to floats in the overrides

This is good to know! I'm not using sensorless on my side and I tested it as far as I could when writing support for it. I'll change that in the next release :)

Tom-T:
I sent a patch for this as I had the same issue. :) I just winged it as I don't really know the language that well.

s0ybean:
I'm still having the same issue even after copying the latest template.  Not sure what it means to "change to floats" in the overrides :)

Can we just keep it simple and use a defined current rather than computing a percentage?

PR #139, opened 2023-04-09
Fixed: Error evaluating 'homing_override:gcode': #137

Tom-T:
I really don't know the language, but it looks like it needs to be set to be a float when assigning the temporary variables. It is now working on my machine without any hacked up solutions. :)


Frix-x:
Thank you for this PR!
You are right, it need to be float :)

PR #145, opened 2023-04-12
Add support for BTT EBB SB2209/SB2240

jakew02:
WIP. Need to implement SPI for 2240 E_STEP


Frix-x (commented):
Thanks for this PR !

I added some comments for some small changes but it's looking almost ok for me now :)

As I currently do no have a solution about the TMC2240 right now. I think we can merge it commented but just add something to specify that it need to be copied in the `overrides.cfg` file for now and not uncommented directly.
I'll try to work on a way to do that :)

Surion79:
Since it is a bigger change for the tmc2240, would it be possible to split this into SB2209 and SB2240? SB2209 shouldn't be as complex as SB2240

Frix-x:
I agree with this. maybe we could just put that in pause until we implement a correct way to handle the different TMC configurations. That way it will be then easier to merge this PR.

jakew02:
Ok I will remove the 2240 stuff for now then On Apr 25, 2023, at 10:15 AM, Félix Boisselier ***@***.***> wrote:﻿
I agree with this. maybe we could just put that in pause until we implement a correct way to handle the different TMC configurations. That way it will be then easier to merge this PR.

Frix-x:
@Surion79, @jakew02, with the work actually merged (#181), do we miss something that can be cherrypicked from this PR? :)

Surion79:
> @Surion79, @jakew02, with the work actually merged (#181), do we miss something that can be cherrypicked from this PR? :)

looked over it, i don't think so.

jakew02:
Yes we can close my PR and go with the other, it’s much more comprehensive!On May 7, 2023, at 5:20 PM, Surion79 ***@***.***> wrote:﻿

@Surion79, @jakew02, with the work actually merged (#181), do we miss something that can be cherrypicked from this PR? :)

looked over it, i don't think so.

Frix-x:
Ok, thank you @jakew02 for your contribution! It helped us 🙂

PR #165, opened 2023-04-27
ercf start print mod

Benoitone:
to be much more compatible with ercf

If using Ercf Happy Hare you must remove in ercf_hardware.cfg (if it had been added previously!!!)

[gcode_macro _USER_VARIABLES]
variable_ercf_enabled: True
gcode:

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -305,9 +305,10 @@ gcode:
     {% if printer.ercf.enable|int == 1 %}
         {% if printer.ercf.tool|int != INITIAL_TOOL or printer.ercf.filament != "Loaded" %}
             ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
-            _DO_ERROR_ERCF_PAUSE
         {% endif %}
-    {% elif ercf_enabled %}
+        _DO_ERROR_ERCF_PAUSE
+    {% endif %}
+    {% if ercf_enabled %}
         ERCF_RESET_STATS
         ERCF_CHANGE_TOOL_STANDALONE TOOL={INITIAL_TOOL}
         _DO_ERROR_ERCF_PAUSE
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -302,12 +302,17 @@ gcode:
     {% endif %}
 
     # Load intial tool if needed
-    {% if ercf_enabled %}
+    {% if printer.ercf.enable|int == 1 %}
+        {% if printer.ercf.tool|int != INITIAL_TOOL or printer.ercf.filament != "Loaded" %}
+            ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
+            _DO_ERROR_ERCF_PAUSE
+        {% endif %}
+    {% elif ercf_enabled %}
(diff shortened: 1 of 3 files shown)


Surion79:
looks good. I hope this fixes some ercf issues i had with klippain :)

Benoitone:
And must remove:

[gcode_macro _USER_VARIABLES]
variable_ercf_enabled: True
gcode:

In HH ercf_hardware.cfg if it had been added previously!!!

Benoitone:
> looks good. I hope this fixes some ercf issues i had with klippain :)

What kind of issues?

Surion79:
The do pause error macro didn't work for me. Yours looks much better.
Currently i don't have that macro

Surion79:
@Benoitone Is this for original ercf or for HappyHare?

Benoitone:
It's for the HH and the one in klippain. But for the one in klippain you must override a lots of parameters.
Personally I didn't use the ercf config in klippain...
I think Moggieuk's HH is actually the best option for the ercf.
For HH, you can install it with its installation script and after you must include in your printer.cfg:
[include ercf_hardware.cfg]
[include ercf_parameters.cfg]
[include ercf_software.cfg]

Surion79:
> 

yeah, but ... ;) I would prefer that the ERCF default is being removed so it does not interfer with HH/confuses users because of double configuration. but thats imho.

Benoitone:
> yeah, but ... ;) I would prefer that the ERCF default is being removed so it does not interfer with HH/confuses users because of double configuration. but thats imho.

If you don’t uncomment the include ercf.cfg in  your klippain printer.cfg it will don’t be load…

Surion79:
> > yeah, but ... ;) I would prefer that the ERCF default is being removed so it does not interfer with HH/confuses users because of double configuration. but thats imho.
> 
> If you don’t uncomment the include ercf.cfg in your klippain printer.cfg it will don’t be load…

i know that :) But why add something that is usually not used? Any issues in the default ercf results in a recommendation to HH.

Benoitone:
> > > yeah, but ... ;) I would prefer that the ERCF default is being removed so it does not interfer with HH/confuses users because of double configuration. but thats imho.
> > 
> > 
> > If you don’t uncomment the include ercf.cfg in your klippain printer.cfg it will don’t be load…
> 
> i know that :) But why add something that is usually not used? Any issues in the default ercf results in a recommendation to HH.

I think because klippain is an evolution of an old config and Frix_x don’t use ercf.
So maybe later klippain ercf config will disappear…

Frix-x:
Hi, so here is what I plan to do in order to satisfy everyone:
  - Modify the original ercf.cfg file from Klippain in order to be more aligned to the original Ette version (maybe also renaming it to ercf_official.cfg)
  - Add a new file for HH named ercf_happy_hare.cfg to add all that is needed for HH.
  
 This will allow the user to select which version he want to use and avoid confusion :)

Surion79:
except the save variables configuration. it should be in a late include so it is the same for alle addons trying to get an own save variables location

Frix-x:
As stated in the other PR from Surion, insted of doing patchs like that. I would like to develop something from scratch and rework the full ERCF system of Klippain. Are you ok with that?

I'll start by creating an issue to discuss about that together before starting :)

PR #169, opened 2023-04-29
added zhop when not homing Z to prevent possible scratching after error while z probing (e.g. tap)

Surion79:
if X or Y is homed without z, it is always checked if a zhop is required. This helps if the probing was errored out, leaving the nozzle on the bed.

### DIFF
--- a/macros/base/homing/homing_overide.cfg
+++ b/macros/base/homing/homing_overide.cfg
@@ -78,6 +78,13 @@ gcode:
 
     G90
 
+    {% if (printer.toolhead.position.z < homing_zhop) %}
+        {% if verbose %}
+            { action_respond_info("Z too low, performing ZHOP") }
+        {% endif %}
+        G0 Z{homing_zhop} F{z_drop_speed}
+    {% endif %}
+
     {% if Z %}
         {% if ('z' in printer.toolhead.homed_axes) %}
             {% if (printer.toolhead.position.z < homing_zhop) %}
--- a/macros/base/homing/homing_overide.cfg
+++ b/macros/base/homing/homing_overide.cfg
@@ -20,8 +20,6 @@ gcode:
     {% set y_position_endstop = printer["configfile"].config["stepper_y"]["position_endstop"]|float %}
     {% set x_position_center = printer.toolhead.axis_maximum.x|int/2 - printer.toolhead.axis_minimum.x|int/2 %}
     {% set y_position_center = printer.toolhead.axis_maximum.y|int/2 - printer.toolhead.axis_minimum.y|int/2 %}
-    
-    {% set homing_force_zhop = printer["gcode_macro _USER_VARIABLES"].homing_force_zhop %}
 
 
     {% if probe_type_enabled == "dockable" %}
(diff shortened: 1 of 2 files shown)


macros/base/homing/homing_overide.cfg
```diff
@@ -77,7 +77,7 @@ gcode:
     {% endif %}
 
     G90
-
+
```
Frix-x, line 80: Be careful you added 4 spaces here. Can you remove them?

Frix-x (changes requested):
Can you remove the spaces added, then I'm ok with this PR to be merged in the next version of Klippain :)

Frix-x (dismissed):

Surion79:
I added a new logic. while it also includes a fix. if Z is NOT requested but homed, it does not check for a required zhop. it just scratches.
The function has to be enabled in the variables. It does not check for previously homed axis. so all acis homing is lost after M84.

macros/base/homing/homing_overide.cfg
```diff
@@ -94,6 +96,22 @@ gcode:
...
+                {% endif %}
+                G0 Z{homing_zhop} F{z_drop_speed}
+            {% endif %}
+        {% elif homing_force_zhop}
+            {% if verbose %}
+                { action_respond_info("Forcing z hop without homing z.") }
+            {% endif %}
+            SET_KINEMATIC_POSITION X=0 Y=0 Z=0
+            G0 Z{homing_zhop} F{z_drop_speed}
+            M84 # Motor off to "unhome" z for functional safety. This will also unhome all previously homed axis
```
Frix-x, line 113: And on this part, I had a look at the documentation and it's perfectly safe to do: `SET_KINEMATIC_POSITION Z=0` without impacting X and Y.
So this means that you can remove the variable and the use of M84. Just do a `SET_KINEMATIC POSITION Z=0` and then the Z hop :)

Also please change the respond_info to : "Homing X or Y while Z is not homed, forcing ZHOP"

user_templates/variables.cfg
```diff
@@ -40,6 +40,10 @@ variable_force_homing_in_start_print: False
 ## Z hop before homing to avoid grinding
 ## the bed due to the gantry sag
 variable_homing_zhop: 5
+## Z hop before homing only X/Y without Z. In case while probing erroring out.
+## For functional safety reasons, if the following option is set to true, 
+## a M84 (Motors off) will be issued between forced Z hop and X/Y homing
+variable_homing_force_zhop: False
```
Frix-x, line 46: With the change asked before, I think this is not needed anymore :)

Frix-x (changes requested):
Hi, sorry for the delay, I wanted to take some time to do it properly.
So I tried to analyse each case and see what would happens. Finally I think this can be made much more simpler without the need of a variable. See my comments :)

Surion79:
@Frix-x I found and even better solution to solve everything. I didn't know that you can use M84 with axis parameter to disable only specific axis. So I changed M84 to M84 Z and did the changes you mentioned

Surion79:
hmm, sad. i saw some links about stating that M84 <axis> works. seems not. need to check why it is documented but not working

macros/base/homing/homing_overide.cfg
```diff
@@ -94,6 +93,13 @@ gcode:
             G0 Z{homing_zhop} F{z_drop_speed}
             {% set X, Y, Z = True, True, True %}
         {% endif %}
+    {% else %}
+        {% if verbose %}
+            { action_respond_info("Homing X or Y while Z is not homed, forcing ZHOP") }
+        {% endif %}
+        SET_KINEMATIC_POSITION Z=0
+        G0 Z{homing_zhop} F{z_drop_speed}
+        M84 Z
```
Frix-x, line 102: I think it's not needed as we know for sure that Z is already not homed

Frix-x (changes requested):

Surion79:
There are two topics in this PR, not just one

macros/base/homing/homing_overide.cfg
Surion79: you described that setting the kinematic position would be like homing manually. this must be undone after the zhop

PR #171, opened 2023-04-29
Park after Flow calibration

Surion79:
added Park and disable Fan after completion of macro


Frix-x (commented):
I'll close this PR as I'm not a fan of modifying the macro to include some custom call like `PARK`.
Indeed, this file is used by people that not necessarly use Klippain directly and so this kind of mods would break it for them.

I have the idea of modifying the helper macro instead to allow this more properly while keeping this untouched as a "part gcode".

PR #176, opened 2023-05-03
Multi TMC support

Frix-x:
Here a new PR about the support of multi TMC based on the work of @Surion79


Frix-x:
As discussed with Surion, I'll merge it now to develop to be able to work on the SB2240 and other toolhead things

PR #177, opened 2023-05-04
Add support for BTT SKR Mini E3 V3 MCU

ndanyluk:
Hi Frix-x,

love the recent update from voron-tools to Klippain :)

Took me a second to get the format, but I added a config for the BTT SKR Mini E3 V3 that I use in my SW.  I think I got it right, but just in case the pinout doc is [here](https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/blob/master/hardware/BTT%20SKR%20MINI%20E3%20V3.0/Hardware/BTT%20E3%20SKR%20MINI%20V3.0_PIN.pdf).

Let me know what you think.  Biggest changes are renaming one of the motor pins I think along with the additional fan2 header.


config/mcu_definitions/BTT_SKR_E3mini_v3.cfg
```diff
@@ -0,0 +1,24 @@
+[board_pins mcu_manufacturer]
+aliases:
+    MCU_XM_STEP=PB13 , MCU_XM_DIR=PB12 , MCU_XM_ENABLE=PB14 ,
+    MCU_YM_STEP=PB10 , MCU_YM_DIR=PB2  , MCU_YM_ENABLE=PB11 ,
+    MCU_ZM_STEP=PB0  , MCU_ZM_DIR=PC5  , MCU_ZM_ENABLE=PB1  ,
+    MCU_EM_STEP=PB3  , MCU_EM_DIR=PB4  , MCU_EM_ENABLE=PD1  ,
+    MCU_TMCUART=PC11 , MCU_TMCTX=PC10  ,
```
Surion79, line 7: MCU_TMCUART and MCU_TMCTX are configured here for all axis. Same pin

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v3.cfg
```diff
@@ -0,0 +1,30 @@
...
+
+[include config/mcu_definitions/BTT_SKR_E3mini_v3.cfg] # Do not remove this line
+[board_pins SKR_mini_e3_mcu]
+mcu: mcu
+aliases:
+    X_STEP=MCU_XM_STEP , X_DIR=MCU_XM_DIR , X_ENABLE=MCU_XM_ENABLE , X_TMCUART=MCU_XM_UART ,
+    Y_STEP=MCU_YM_STEP , Y_DIR=MCU_YM_DIR , Y_ENABLE=MCU_YM_ENABLE , Y_TMCUART=MCU_YM_UART ,
+    Z_STEP=MCU_ZM_STEP , Z_DIR=MCU_ZM_DIR , Z_ENABLE=MCU_ZM_ENABLE , Z_TMCUART=MCU_XM_UART ,
+    E_STEP=MCU_EM_STEP , E_DIR=MCU_EM_DIR , E_ENABLE=MCU_EM_ENABLE , E_TMCUART=MCU_EM_UART ,
+
```
Surion79, line 19: You are missing here the UART and TX of the other file i commented.

Surion79:
Have you a running printer with this kind of hardware configuration? Otherwise I have a running SKR Pico, which is similar to the mini regarding uart/tx, which I plan to PR this weekend.

config/mcu_definitions/BTT_SKR_E3mini_v3.cfg
ndanyluk: Hi, I think this is correct and same as how the SKR Mini E3 v2 handles it.  All of the drivers are tied to the same uart line I think

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v3.cfg
ndanyluk: Hi @Surion79 sorry if those are missing I adapted the existing SKR Mini E3 v2 config already in the repo for this board revision.  I'll take a look and revise if needed

config/mcu_definitions/BTT_SKR_E3mini_v3.cfg
Surion79: i commented it as reference to my other comment :)

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v3.cfg
Surion79: the v2 is unfortunately not (longer) working. assigning multiple aliases to same pin results in an error. The uart and tx must be implemented by overriding the tmc config by adding it in the overrides.cfg. I'll get a PR for a working version read and update the v2 accordingly. what changes are between v2 and v3?

ndanyluk: @Surion79 got it - I didn't realize that.  The only changes from V2 to V3 are the changing of some pin assignments (I think on one of the motor pins) and the addition of a FAN2 port that is typically used for controller fans, but can be for anything.  I don't mind updating my config - do you have an example I can follow?

ndanyluk:
@Surion79 I'll test this in a minute, but before you start reworking configs - I added a "shared UART" config option for the motors - still uses the default wiring files, but added the necessary tx_pin and uart_address parameters.  Let me check it out and I'll get back to you

Surion79:
> Well this sucks - I'm stuck at an apparent limit of the board_aliases function. It seems that an alias can't be used to define the UART address. getting "Unable to parse option 'uart_address' in section 'tmc2209 stepper_x'"

uart_address is an integer, not a pin. Not sure to what you are referring

Surion79:
@ndanyluk You can look up my new PR regarding SKR Pico. This is copied from my running v0 with modified klippain. So it is not an academic exercice :D

ndanyluk:
@Surion79 ooh I like that - will rebuild this to follow your approach tonight.  Might also take a stab at applying that to the SKR Mini E3 V2 config as well

Surion79:
> @Surion79 ooh I like that - will rebuild this to follow your approach tonight. Might also take a stab at applying that to the SKR Mini E3 V2 config as well

that would be great :)

ndanyluk:
Ok, I got eager, haven't applied the update yet on my printer, but modified the files to follow your lead.  So I can test, where do you intend the overrides for the Pico to get applied? in the overrides.cfg or in the MCU.cfg?

Surion79:
Overrides.cfg is the correct place

ndanyluk:
@Surion79 one idea I just had and I'm wondering your thoughts on: why don't we put the tmc settings on the MCU template itself? 

At least for these boards, it's baked into the board itself (aka these drivers aren't user replaceable so they should never change).  The only place these parameters are called out are in 'default_wiring.cfg' which mentions the uart pin & this new override section.

If we instead put it in the MCU default template, then no need for a new function to inject this "mandatory override" in the installer (although I can see moving the uart pin to the MCU config becoming a breaking change).  

The MCU templates already prescribe defaults for which motor is considered X, Y,  etc on the Octopus and other boards, this feels to me like the right place to put this based on that philosophy.

What do you think?

Surion79:
Interesting idea, although you can't 'undo' includes. With your idea all defined tmc must have a tx pin defined and an uart_address. You can't 'unoverride' extruder tmc of a mcu toolhead like the picobilical or other can bus toolheads when it is defined in mcu.cfg.
Overrides should be in a single place and not ripped apart. (E.g., Xyz in mcu.cfg and e in overrides.cfg)
That's imho

ndanyluk:
> Interesting idea, although you can't 'undo' includes. With your idea all defined tmc must have a tx pin defined and an uart_address. You can't 'unoverride' extruder tmc of a mcu toolhead like the picobilical or other can bus toolheads when it is defined in mcu.cfg. Overrides should be in a single place and not ripped apart. (E.g., Xyz in mcu.cfg and e in overrides.cfg) That's imho

Ahh yeah I didn't consider that.  Okay fair enough then! That would be a real pain down the road, especially as CAN boards & the Picobilical in particular become way more common.

ndanyluk:
Ok, I think I cleaned it up so the merge conflicts with develop branch are gone.  Will test in the morning

Frix-x:
> @Surion79 one idea I just had and I'm wondering your thoughts on: why don't we put the tmc uart settings on the MCU template itself?
> 
> At least for these boards, it's baked into the board itself (aka these drivers aren't user replaceable so they should never change). The only place these parameters are called out are in 'default_wiring.cfg' which mentions the uart pin & this new override section.
> 
> If we instead put it in the MCU default template, then no need for a new function to inject this "mandatory override" in the installer (although I can see moving the uart pin to the MCU config becoming a breaking change).
> 
> The MCU templates already prescribe defaults for which motor is considered X, Y, etc on the Octopus and other boards, this feels to me like the right place to put this based on that philosophy.
> 
> What do you think?

I'm in for that. I was trying to think about how to handle those specific board overrides correctly and I think this is the easiest. They are "static" overrides that go with the boards so I think they can be put directly in the mcu.cfg file :)

user_templates/mcu_defaults/main/BTT_SKR_Mini_E3_v3.cfg
```diff
@@ -27,4 +29,4 @@ aliases:
     PART_FAN=MCU_FAN0 , E_FAN=MCU_FAN1 , CONTROLLER_FAN=MCU_FAN2 ,
 
     STATUS_NEOPIXEL=MCU_NEOPIXEL ,
-    LIGHT_NEOPIXEL=MCU_NEOPIXEL  ,
+    LIGHT_NEOPIXEL=MCU_NEOPIXEL  ,
```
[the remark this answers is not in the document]
ndanyluk: thanks, I didn't realize this was double-assigned.  The BTT SKR Mini E3 V2 config does the same thing and I built off that.  Will remove shortly

Surion79:
> 

this only works for motors you usually use. meaning that e.g. Extruder is not something for mcu.cfg in this case? needs to be defined what goes where. if users don't know where to look anymore, it gets confusing and users probably just past everything in overrides.cfg to have a ONE "go-to" place for individual configuration

Frix-x:
Hey, now that we have merged the new structure for the TMC system and also have simplified the "static toolhead overrides" for pin remaping, can you update the PR for a merge? :)

config/mcu_definitions/BTT_SKR_E3mini_v3.cfg
ndanyluk: resolved per other discussions

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v3.cfg
ndanyluk: resolving conversation per our other discussions

user_templates/mcu_defaults/main/BTT_SKR_Mini_E3_v3.cfg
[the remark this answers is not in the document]
ndanyluk: fixed in latest commit

Frix-x:
Thanks! I'll merge it :)

ndanyluk:
@Frix-x thanks! Sorry I didn't get a chance to squash the commit history to make it cleaner.

btw, I did test this on my VSW (non-motion because of other missing variables related to CoreXZ) and confirmed that this config does work:
- temperature sensors appear as expected
- heaters work as expected

However, I did catch one last issue that didn't make it into my last commit :(
- the include in the mcu_default config is wrong - needs to use the main folder now.  Will open another PR to fix

PR #179, opened 2023-05-04
Remove "position_endstop: 0" from Trident Z config files.

jurikaur:
Klipper errors out when activating Voron Tap probe.


jurikaur:
sorry, found this update in "develop" branch.

PR #181, opened 2023-05-04
Config toolheads sb2209/ sb2240

Surion79:
initial version for review
added SB2209 and SB2240 configuration including SB2240 override file. 
segmented configs in different folders as agreed previously

Due to the missing possibility to multi-assign aliases to a shared pin, it is not possible to do a default config without overwrite. The ADXL Aliases are kept, and the related TMC SPI config has been overriden to use the correct pins.


jurikaur:
Good I have to try it.

ndanyluk:
@Surion79 just a thought echoed from my own PR:

I think it might be easier to maintain if the overrides.cfg is merged into the mcu_template.  I know the template currently only has [board_aliases] but that doesn't need to be the only section, and this would avoid needing to modify the installer to point to this override file.

Also at an conceptual level, overrides.cfg is all about customization, right? For something static that should be the default, I feel that the template file is a good place for it, and then someone can write their own overrides.cfg from there if they want to adjust.

Just my take on it.  I'm new to this project since it became klippain and I don't know what prior discussions were had on this.

ndanyluk:
> @Surion79 just a thought echoed from my own PR:
> 
> I think it might be easier to maintain if the overrides.cfg is merged into the mcu_template. I know the template currently only has [board_aliases] but that doesn't need to be the only section, and this would avoid needing to modify the installer to point to this override file.
> 
> Also at an conceptual level, overrides.cfg is all about customization, right? For something static that should be the default, I feel that the template file is a good place for it, and then someone can write their own overrides.cfg from there if they want to adjust.
> 
> Just my take on it. I'm new to this project since it became klippain and I don't know what prior discussions were had on this.

Per our discussion in #177, there's more to the overrides vs mcu template than I had considered at a klipper level (ability to "undo includes" poses a particular challenge).  Maybe there will be a better way to handle in the future but for now this is a solid approach.

I'm going to pick this up in my fork and test on my Switchwire (as soon as I can get my BTT SKR Mini E3 working first)

Surion79:
The best way would be first to get rid of the issue not assigning multiple aliases to the same pin. It would make it much less painful.
But thats a Klipper topic

Frix-x:
> I think it might be easier to maintain if the overrides.cfg is merged into the mcu_template. I know the template currently only has [board_aliases] but that doesn't need to be the only section, and this would avoid needing to modify the installer to point to this override file.

Thanks for the suggestion and this is a good idea that I did not thought about it bedore now. I think we will do that to make it more clean and easy for the installer !
Regarding "undoing" an include, this is a tricky point that is not possible. However, nothing will prevent us to just pre-comment something and let the user uncomment if needed (rather than including by default and having trouble later)

Surion79:
a clear guideline must be established. if a user needs to edit many different files to get the printer going, it begins to be more hassle than just using a printer provided template (e.g. Voron) and just modify it.

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
```diff
@@ -0,0 +1,22 @@
+[board_pins toolhead_manufacturer]
+mcu: toolhead
+aliases:
+   MCU_TMCDRIVER_STEP=PD0 , MCU_TMCDRIVER_DIR=PD1 , MCU_TMCDRIVER_ENABLE=PD2 , MCU_TMCDRIVER_UART=PA15 ,
+   MCU_TMCDRIVER_DIAG=PB3 ,
```
Frix-x, line 5: As per the documentation (see here: https://cdn.shopify.com/s/files/1/1619/4791/t/48/assets/-8-1676011507868.jpg?v=1676011516)

I can see that motor pins are named "MOTOR Drive" and with some slight change also to the different pins names. So I would suggest using something like: `MCU_MOTORDRIVE_STEP`, `MCU_MOTORDRIVE_DIR`, `MCU_MOTORDRIVE_EN`, `MCU_MOTORDRIVE_UART` and `MCU_MOTORDRIVE_DIAG`

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
```diff
@@ -0,0 +1,22 @@
...
+
+   MCU_ENDSTOP1=PB6 , MCU_ENDSTOP2=PB5 , MCU_ENDSTOP3=PB7 ,
```
Frix-x, line 7: Endstops should be called `MCU_STOP1`, `MCU_STOP2` and `MCU_STOP3`

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
```diff
@@ -0,0 +1,22 @@
...
+   MCU_PROBE=PB8 ,
+   MCU_SERVOS=PB9 ,
+
+   MCU_HOTEND0=PB13 ,
```
Frix-x, line 11: This is `MCU_HE0`

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
```diff
@@ -0,0 +1,22 @@
...
+   MCU_TH0=PA3 ,
+   
+   MCU_FAN1=PA0 , MCU_FAN2=PA1 ,
```
Frix-x, line 14: This is apparently `MCU_FAN1_PWM` and `MCU_FAN2_PWM` (I don't really like it but it's what is in the doc)

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
Frix-x, line 1: I think you can also add `MCU_NTC=PA2` as it's present in the doc for this board

config/mcu_definitions/toolhead/BTT_SB2240_v1.0.cfg
```diff
@@ -0,0 +1,21 @@
+[board_pins toolhead_manufacturer]
+mcu: toolhead
+aliases:
+   MCU_MOTOR_SPI_NSS=PA15 , # CLK/MOSI/MISO are shared with SPI2 configuration
+   
+   MCU_ENDSTOP1=PB6 , MCU_ENDSTOP2=PB5 , MCU_ENDSTOP3=PB7 ,
```
Frix-x, line 6: Endstops should be called `MCU_STOP1`, `MCU_STOP2` and `MCU_STOP3`

config/mcu_definitions/toolhead/BTT_SB2240_v1.0.cfg
```diff
@@ -0,0 +1,21 @@
...
+   MCU_PROBE=PB8 ,
+   MCU_SERVOS=PB9 ,
+
+   MCU_HOTEND0=PB13 ,
```
Frix-x, line 10: This is `MCU_HE0`

config/mcu_definitions/toolhead/BTT_SB2240_v1.0.cfg
```diff
@@ -0,0 +1,21 @@
...
+   MCU_TH0=PA3 ,
+   
+   MCU_FAN1=PA0 , MCU_FAN2=PA1 ,
```
Frix-x, line 13: This is apparently `MCU_FAN1_PWM` and `MCU_FAN2_PWM` (I don't really like it but it's what is in the doc)

config/mcu_definitions/toolhead/BTT_SB2240_v1.0.cfg
```diff
@@ -0,0 +1,21 @@
+[board_pins toolhead_manufacturer]
+mcu: toolhead
+aliases:
+   MCU_MOTOR_SPI_NSS=PA15 , # CLK/MOSI/MISO are shared with SPI2 configuration
```
Frix-x, line 4: On this board, there is also UART available so we can add it to the manufacturer file I think: `MCU_MOTORDRIVE_STEP=PD0`, `MCU_MOTORDRIVE_DIR=PD1`, `MCU_MOTORDRIVE_EN=PD2` and `MCU_MOTORDRIVE_DIAG=PB3`

user_templates/mcu_defaults/toolhead/BTT_SB2209_v1.0.cfg
```diff
@@ -0,0 +1,33 @@
...
+    TOOLHEAD_SENSOR=MCU_ENDSTOP3 ,
+
+    E_HEATER=MCU_HOTEND0 , E_TEMPERATURE=MCU_TH0 ,
+
+    PART_FAN=MCU_FAN1 , E_FAN=MCU_FAN2 ,
+
+    STATUS_NEOPIXEL=MCU_RGB ,
+
+    ADXL_CS=MCU_SPI2_CS , ADXL_SCLK=MCU_SPI2_SCLK , ADXL_MISO=MCU_SPI2_MISO , ADXL_MOSI=MCU_SPI2_MOSI ,
+    MAX31865_NSS=MCU_SPI1_NSS , MAX31865_CLK=MCU_SPI1_CLK , MAX31865_MOSI=MCU_SPI1_MOSI , MAX31865_MISO = MCU_SPI1_MISO ,
```
Frix-x, line 32: Please change the aliases accordingly to the manufacturer files

user_templates/mcu_defaults/toolhead/BTT_SB2240_v1.0.cfg
```diff
@@ -0,0 +1,33 @@
...
+    TOOLHEAD_SENSOR=MCU_ENDSTOP3 ,
+
+    E_HEATER=MCU_HOTEND0 , E_TEMPERATURE=MCU_TH0 ,
+
+    PART_FAN=MCU_FAN1 , E_FAN=MCU_FAN2 ,
+
+    STATUS_NEOPIXEL=MCU_RGB ,
+
+    ADXL_CS=MCU_SPI2_CS , ADXL_SCLK=MCU_SPI2_SCLK , ADXL_MISO=MCU_SPI2_MISO , ADXL_MOSI=MCU_SPI2_MOSI ,
+    MAX31865_NSS=MCU_SPI1_NSS , MAX31865_CLK=MCU_SPI1_CLK , MAX31865_MOSI=MCU_SPI1_MOSI , MAX31865_MISO = MCU_SPI1_MISO ,
```
Frix-x, line 32: Please change the aliases accordingly to the manufacturer files

user_templates/mcu_defaults/toolhead/overrides/BTT_SB2240_v1.0_overrides.cfg
```diff
@@ -0,0 +1,15 @@
...
+# toolhead MCU during the installation process. They should provide a
+# good base to work with. Feel free to adapt to your board if needed!
+
+
+# Since ADXL shares pins with the SPI interface of the TMC, it must be set up here, since different pin aliases must not resolve to the same pin
+[tmc2240 extruder]
+cs_pin: toolhead:MCU_MOTOR_SPI_NSS
+spi_software_sclk_pin: toolhead:ADXL_SCLK
+spi_software_mosi_pin: toolhead:ADXL_MOSI
+spi_software_miso_pin: toolhead:ADXL_MISO
```
Frix-x, line 15: As per discussion in other PRs, as this is some pin mappings, please add them directly to the MCU file. You can also add the default overrides that you can find in this file as well: https://github.com/Frix-x/klippain/blob/develop/user_templates/mcu_defaults/toolhead/overrides/default.cfg

Frix-x (changes requested):
Thanks for this PR. I think it's already pretty good and almost ready to be merged.

Can you just do the changes mainly to follow more closely the manufacturer docs?

config/mcu_definitions/toolhead/BTT_SB2209_v1.0.cfg
Surion79: For archive sake, i highly suggest using original sources of the document and not links to a webshop CDN.

Surion79:
i reworked the branch so the tmc sections for toolheads moved from default to the mcu-related sections. otherwise it would conflict toolheads with !tmc2209

config/mcu_definitions/toolhead/BTT_SB2240_v1.0.cfg
Surion79: and where is the UART pin? 
BTT only defines the tmc2240 for SPI in their config. So i'd rather not include something that is not meant to be used.

Frix-x: Yeah ok no problem

Frix-x (approved):

PR #184, opened 2023-05-06
 Added tap safety check and updated print_start macro

jurikaur:
1. added TAP safety code
2. updated TAP startprint action sequence
3. added start_print new _MODULE for preheating extruder to tap
   temperature before probing.

### DIFF
--- a/config/hardware/probes/tap_probe.cfg
+++ b/config/hardware/probes/tap_probe.cfg
@@ -1,6 +1,6 @@
 [gcode_macro _USER_VARIABLES]
 variable_probe_type_enabled: "vorontap"
-variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "tap_heating", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
+variable_startprint_actions: "bed_soak", "extruder_preheating", "chamber_soak", "clean", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
 gcode:
 
 # TAP probe definition also include the probe management macro directly from here
@@ -18,23 +18,6 @@ samples_result: median
 sample_retract_dist: 1.0
 samples_tolerance: 0.005
 samples_tolerance_retries: 3
-# Add safety feature before probing bed
-activate_gcode:
-    {% set PROBE_TEMP = printer["gcode_macro _USER_VARIABLES"].tap_max_probing_temp|float %}
-    {% set MAX_TEMP = PROBE_TEMP + 5 %}
-    {% set ACTUAL_TEMP = printer.extruder.temperature %}
-    {% set TARGET_TEMP = printer.extruder.target %}
-
-    {% if TARGET_TEMP > PROBE_TEMP %}
-        { action_respond_info('Extruder temperature target of %.1fC is too high, lowering to %.1fC' % (TARGET_TEMP, PROBE_TEMP)) }
-        M109 S{ PROBE_TEMP }
-    {% else %}
-        # Temperature target is already low enough, but nozzle may still be too hot.
-        {% if ACTUAL_TEMP > MAX_TEMP %}
-            { action_respond_info('Extruder temperature %.1fC is still too high, waiting until below %.1fC' % (ACTUAL_TEMP, MAX_TEMP)) }
-            TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={ MAX_TEMP }
-        {% endif %}
-    {% endif %}
 
 [stepper_z]
 endstop_pin: probe:z_virtual_endstop
--- a/config/hardware/probes/tap_probe.cfg
+++ b/config/hardware/probes/tap_probe.cfg
@@ -1,6 +1,6 @@
 [gcode_macro _USER_VARIABLES]
 variable_probe_type_enabled: "vorontap"
-variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
+variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "tap_heating", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
 gcode:
 
 # TAP probe definition also include the probe management macro directly from here
@@ -18,6 +18,23 @@ samples_result: median
 sample_retract_dist: 1.0
 samples_tolerance: 0.005
 samples_tolerance_retries: 3
+# Add safety feature before probing bed
+activate_gcode:
+    {% set PROBE_TEMP = printer["gcode_macro _USER_VARIABLES"].tap_max_probing_temp|float %}
+    {% set MAX_TEMP = PROBE_TEMP + 5 %}
+    {% set ACTUAL_TEMP = printer.extruder.temperature %}
+    {% set TARGET_TEMP = printer.extruder.target %}
+
+    {% if TARGET_TEMP > PROBE_TEMP %}
+        { action_respond_info('Extruder temperature target of %.1fC is too high, lowering to %.1fC' % (TARGET_TEMP, PROBE_TEMP)) }
+        M109 S{ PROBE_TEMP }
+    {% else %}
+        # Temperature target is already low enough, but nozzle may still be too hot.
+        {% if ACTUAL_TEMP > MAX_TEMP %}
+            { action_respond_info('Extruder temperature %.1fC is still too high, waiting until below %.1fC' % (ACTUAL_TEMP, MAX_TEMP)) }
+            TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={ MAX_TEMP }
+        {% endif %}
+    {% endif %}
 
 [stepper_z]
 endstop_pin: probe:z_virtual_endstop
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -117,8 +117,8 @@ gcode:
             _MODULE_BED_MESH
         {% elif action == "primeline" %}
             _MODULE_PRIMELINE
-        {% elif action == "custom1" %}
-            _MODULE_CUSTOM1
+        {% elif action == "tap_heating" %}
+            _MODULE_EXTRUDER_HEATING_TAP
         {% elif action == "custom2" %}
             _MODULE_CUSTOM2
         {% elif action == "custom3" %}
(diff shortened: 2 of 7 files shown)


Surion79:
you found it? (:

jurikaur:
I did it in wrong order, synced develop branch after my own push. I trie to do it oposite order

jurikaur:
Don't know how to revert sync-commit order. If you know how, please let me know.

Surion79:
the custom1,2,3 are meant for personal additions, not PRs on klippain. they should be individual

Surion79:
but, is tap heating required?

jurikaur:
> the custom1,2,3 are meant for personal additions, not PRs on klippain. they should be individual

Ok, I will correct that.

jurikaur:
> but, is tap heating required?

Heating is reqiured to soften filament remnants on nozzle so it wont affect probebing accuracy. Also to get extruder warm and heat expanded.

Frix-x:
Hello, thanks for this PR.

Did you already used the default procedure? Because in fact the system in your PR is already something managed automatically under the hood in Klippain. I'm pretty sure there is no need for this.

Since we support multiple probe types in Klippain, the activation/deactivation procedure for each of them are managed in https://github.com/Frix-x/klippain/blob/main/macros/base/probing/generic_probe.cfg
You will find proper TAP temperature management directly in this macro :)

Surion79:
@jurikaur is rights, @Frix-x : Klippain does not heat the tap to 150 degrees. it only cools down if hotter. currently starting a print and it is probing at 0 degree set for heating.

Benoitone:
It seems to me that it's already included in klippain...

Surion79:
> It seems to me that it's already included in klippain...

nope. he wants initial heating of tap which is not part of klippain

Surion79:
somewhere for tap it requires a
`M109 S150`

macros/base/start_print.cfg
```diff
@@ -403,14 +403,39 @@ gcode:
...
-    # gcode:
-    #   ## Your custom code here
+    # Heat the nozzle to tap probing temperature ontop of the purge bucket (if available)
+    # or in the middle of the bed and wait 30s to let the nozzle dilate and stabilize
+    {% set EXTRUDER_TEMP = printer["gcode_macro _USER_VARIABLES"].tap_max_probing_temp|float %}
+    {% set INITIAL_TOOL = printer["gcode_macro START_PRINT"].initial_tool %}
+
+    {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
+    {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
+    {% set ercf_enabled = printer["gcode_macro _USER_VARIABLES"].ercf_enabled %}
```
Surion79, line 415: for what do you need ercf here?

Surion79:
@jurikaur I am not so happy that you did multiple things in a module. there is already a clean module. i'd rather have a module to exactly what it is describing. heating up nozzle for tap

config/hardware/probes/tap_probe.cfg
```diff
@@ -1,6 +1,6 @@
 [gcode_macro _USER_VARIABLES]
 variable_probe_type_enabled: "vorontap"
-variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
+variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "tap_heating", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
```
Frix-x, line 3: Can you change the name of the module to "extruder_preheating" to add genericity as it could be useful for other kind of probes also.

macros/base/start_print.cfg
```diff
@@ -117,8 +117,8 @@ gcode:
             _MODULE_BED_MESH
         {% elif action == "primeline" %}
             _MODULE_PRIMELINE
-        {% elif action == "custom1" %}
-            _MODULE_CUSTOM1
+        {% elif action == "tap_heating" %}
```
Frix-x, line 120: Can you change the name of the module to "extruder_preheating" to add genericity as it could be useful for other kind of probes also.

config/hardware/probes/tap_probe.cfg
```diff
@@ -18,6 +18,23 @@ samples_result: median
...
+    {% if TARGET_TEMP > PROBE_TEMP %}
+        { action_respond_info('Extruder temperature target of %.1fC is too high, lowering to %.1fC' % (TARGET_TEMP, PROBE_TEMP)) }
+        M109 S{ PROBE_TEMP }
+    {% else %}
+        # Temperature target is already low enough, but nozzle may still be too hot.
+        {% if ACTUAL_TEMP > MAX_TEMP %}
+            { action_respond_info('Extruder temperature %.1fC is still too high, waiting until below %.1fC' % (ACTUAL_TEMP, MAX_TEMP)) }
+            TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={ MAX_TEMP }
+        {% endif %}
+    {% endif %}
```
Frix-x, line 37: This can be removed as already managed by https://github.com/Frix-x/klippain/blob/main/macros/base/probing/generic_probe.cfg

macros/base/start_print.cfg
```diff
@@ -403,14 +403,39 @@ gcode:
...
 gcode:
-    # ---- CUSTOM Macro section
-    # this section is reserved for personal customized start actions, which can be combined with all other start actions
-    # in order to use this, create a new macro in overrides.cfg
-    # [gcode_macro _MODULE_CUSTOM1]
-    # gcode:
-    #   ## Your custom code here
+    # Heat the nozzle to tap probing temperature ontop of the purge bucket (if available)
+    # or in the middle of the bed and wait 30s to let the nozzle dilate and stabilize
+    {% set EXTRUDER_TEMP = printer["gcode_macro _USER_VARIABLES"].tap_max_probing_temp|float %}
```
Frix-x, line 410: This should be changed to use the safe extruder temp instead:
```
```

macros/base/start_print.cfg
```diff
@@ -403,14 +403,39 @@ gcode:
...
+    {% if status_leds_enabled %}
+        STATUS_LEDS COLOR="HEATING"
+    {% endif %}
+
+    _CONDITIONAL_MOVE_TO_PURGE_BUCKET
+    {% if not purge_and_brush_enabled %}
+        G0 X{max_x|int / 2} Y{max_y|int / 3} Z50 F{St}
+    {% endif %}
+
+    M109 S{EXTRUDER_TEMP}
```
Frix-x, line 434: Since the safe extruder temp is usually a temperature that doesn't allow material to flow properly, I would suggest to only heat to this temperature and do not care for the purge bucket position, etc... in this module.

Keep the status leds heating color, keep the M109 and basically that's it

Frix-x (changes requested):
Mhmm Ok. I understand the problem and why it's working correctly for most of the people. There is indeed already a specific line for extruder preheating at a safe temperature here: https://github.com/Frix-x/klippain/blob/52632ec5cd482282c1319fd6d01a7499305d2a1e/macros/base/start_print.cfg#L246-L247

But this is part of the chamber heating module and happens only if you have a chamber temperature sensor and a CHAMBER parameter set in your PRINT_START from the slicer. So I agree that this could be reworked using a new module.

So @jurikaur can you modify your PR accordingly?

Also I think if you accept to go this route, you can safely remove the safe temperature lines from the chamber heating module and update all the probe `variable_startprint_actions` to add your new preheating directly before the chamber heating :)

macros/base/start_print.cfg
Frix-x: I added a proper and full review, So indeed this line is not needed, but there is more to remove. Please see below (I'll solve this conversation to keep thing in the full review and avoid dispersion :) )

Surion79:
i really like this kind of solution :)

config/hardware/probes/tap_probe.cfg
jurikaur: Renamed as suggested

macros/base/start_print.cfg
jurikaur: Renamed as requested. Added back Custom1 part.

config/hardware/probes/tap_probe.cfg
jurikaur: Removed this code part.

macros/base/start_print.cfg
jurikaur: Updated variable.

macros/base/start_print.cfg
jurikaur: Updated macro M109 S{safe_extruder_temp}

config/hardware/probes/tap_probe.cfg
```diff
@@ -1,6 +1,6 @@
 [gcode_macro _USER_VARIABLES]
 variable_probe_type_enabled: "vorontap"
-variable_startprint_actions: "bed_soak", "chamber_soak", "clean", "extruder_preheating", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
+variable_startprint_actions: "extruder_preheating", "bed_soak", "chamber_soak", "clean", "tilt_calib", "z_offset", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
```
Surion79, line 3: i suggest to do the extruder preheating after bed_soak, as the bed soak takes the longest amount of time. no need to waste power on the extruder heating while waiting for bed soak

Frix-x: Mhmm I'm open to move it where you prefer :)

  - First position allow to heat the chamber quickly and save time as we do it in parallel to the bed soak but this is not always needed
  - After bed soak (but before chamber soak) make it more generic if no chamber soak is needed and allow to save some energy but can make the process a little bit longer if a chamber soak is needed

So ok, let's go for more genericity by putting it in second position. Users will be able override it if they want to put it in front anyway.

Frix-x (changes requested):
@jurikaur I think we are almost done for a merge with this PR :)
Can you do the last two small change?

Also, may I ask you to update:
  1. The other probes cfg files to also include this preheating natively before the chamber heating for every of them
  2. Remove this line from the chamber heating module as it's not needed anymore: https://github.com/Frix-x/klippain/blob/5d68624b815eb1b60cdccdf00e667eaaa33f0f8a/macros/base/start_print.cfg#L255-L256 You can also update the comment L232 and remove the variable L238 as it's not needed anymore as well.
  3. Please also update this to add your new module in the documentation: https://github.com/Frix-x/klippain/blob/5d68624b815eb1b60cdccdf00e667eaaa33f0f8a/user_templates/overrides.cfg#L29

jurikaur:
I updated my code as suggested, please review. If it is ok I will squash my commits

Frix-x (approved):
Yes, this is good for me now! Thanks you for this addition :)

jurikaur:
I will rework this, I do not know how it is going to end. Learning. Maybe I have to create new PR.

Frix-x:
No problem, take your time :)

PR #186, opened 2023-05-07
Deleted old "tap_heating" from voron tap config

jurikaur:
Somehow missed this when doing patch rework.


Frix-x (approved):
Oops, did not see it also. I'll merge it now

PR #188, opened 2023-05-08
Klippain v4

Frix-x:
Here is the PR for Klippain v4 that is almost ready.

Klippain v4 is packed with a lot of new features and documentation !

## New features and changes

- Reworked all the Klippain install and config documentation to make it more user-friendly! (275adcb11481e58deb4df5b9349f5cea73973850 182c667e0fa607bb3d61d0daa408fc31970da623 fe8216102a4d917bfbe1b2888d35cff336521512)
- Added full input shaper results interpretation documentation to help you understand the different graphs and tune your machine (2c41a777cc9a42c5ba481bbba610ba8490e7287c f4ce997f0ab0519bced651c08eb50833732dad9d)
- Added support for multi-kinematics: Klippain can now run on Cartesian or CoreXZ printers while also supporting virtual probes types! (f20f395cdc747a80735daf7d0bcc7049f1374518 e2e84081249b78737e50daee65bcc6accd8ac025 7a93165eb74cbb0577288b3b72713021b561541b)
  > **Warning**:
  > 
  > **Breaking change**: you will need to get [the new printer.cfg base template](https://github.com/Frix-x/klippain/blob/9493ce69fd8dc8f03ed8b0372e79dbd8bd2b02e7/user_templates/printer.cfg) to get the new kinematics and virtual probes includes available. Also please look at the whole file as there was a lot of refactoring to allow the multi-kinematics and some of the includes changed their names an paths.
- This also allowed the addition of multi-TMC support to allow selecting a different TMC for every axis and extruder!  (333d802f2fe6108a3f03b12d540f68c20bc66c74 e2e84081249b78737e50daee65bcc6accd8ac025)
  > **Warning**:
  > 
  > **Breaking change**: you will need to get [the new mcu.cfg base template](https://github.com/Frix-x/klippain/blob/9493ce69fd8dc8f03ed8b0372e79dbd8bd2b02e7/user_templates/mcu.cfg) to get the new TMC includes available. Also you should update the MCU templates corresponding to your boards as the `[include config/mcu_definitions/... ]` line has changed.
- Added an example of Z motor direction override directly in the `overrides.cfg` user template as it's one of the most used override (f010c653f53d2a1f01de3aec83bfcdab7b598bbf)
- Commented the default PID overrides and added some instructions about it to make it more easy to understand (d98b1508b792321c0a86ac6b426fac9f84edb77e)
- Custom `START_PRINT` modules are available for advanced users that want to do some "custom actions" in their `START_PRINT` routine (9d74301db989539fde9d5b4b87d6181cd2ba8ff7 82ef7c292c5dfe88cd5e2691d993c23a3e4f8574 b17139f7a416fed971fb925b64460a99bc127974)
- Added Orbiter2.0 extruder support (27de9b0cf284bb7a82058ad75ca4838b2ef83773 6cc9dcb4a6120a47f8ddabedb15e1a0544d3fc5d)
- The prime line has been modified to be more efficient. Now it's smaller and thicker and it makes it easier to remove the extruded plastic from the bed. Another point is that it allows it to work on V0 and small beds. Finally, everything is easy to set up and Klippain will take care of the firmware limits to adjust the prime line parameters on the fly to avoid errors or to make sure the prime line sticks to the bed and is not too thin! (b2916dc05228f138ccf619eedcd8ce923de8a8c0 ec9a581c4d35ca174b83c5498512cdb5a80d723f)
  > **Warning**:
  > 
  > **Breaking change**: you need to add or update some variables in your `variables.cfg` file:  https://github.com/Frix-x/klippain/blob/1539dc4eb5f026698f48925398e910dbf5b03c54/user_templates/variables.cfg#L57-L62
- Updated Marlin compatibility macros and added an M900 macro mapped to Klipper pressure advance (5d68624b815eb1b60cdccdf00e667eaaa33f0f8a)
- Full rework of the ERCF integration. Now Klippain is natively compatible with the [HappyHare](https://github.com/moggieuk/ERCF-Software-V3) software backend (1ea04e34ff3b0ae4dd033d59cfcdb81abbebdc60)
  > **Warning**:
  > 
  > **Breaking change**: you will need to get [the new printer.cfg template](https://github.com/Frix-x/klippain/blob/f46dea7b7da3987375f4086321b581d82b3878fc/user_templates/printer.cfg) to get the new ERCF includes available. Also you will need to add some new variables to your `variables.cfg` file:  https://github.com/Frix-x/klippain/blob/1ea04e34ff3b0ae4dd033d59cfcdb81abbebdc60/user_templates/variables.cfg#L174-L181
- Added an "extruder pre-heating" module to choose when to do it in the `START_PRINT` sequence (ed9a2939a22d3b5502438b343ad421c33a519ad9 16013f4a73b0dcce65e778f22946d77ba533a18b f46dea7b7da3987375f4086321b581d82b3878fc)
- Added support for new toolhead boards: BTT SB2209 and BTTSB2240 (41dee4b88faab5786371cb8ce895ffa94ae9a155 b6f36a535d59f681e4269d7036f3400383f9e67d 25e0dba36bde366badfa86191fad3865b5b5f962 2e6b912004811e0e55e2758c3711d459c432a5b6 f783b29dda1e499a3281e7095f22f85a410b7801 2f48c585e0b9cda1138923b00eca047762eaaae3)
- Added support for BTT SKR Pico v1.0 MCU (c965fd707632030273b4f1b116ec48420d74b15e)
- Added support for BTT SKR E3 mini v3 MCU (f68abff 1ffd2130e4543418ca5e9c2d4914db471e015029)
- Refactored the BTT SKR Mini E3 boards name to be more in line with the maket product name (3f65109)
  > **Warning**:
  > 
  > **Breaking change**: if already using the BTT SKR Mini E3 v2 MCU, you will need to update your mcu.cfg file by adding [the new SKR E3 mini template](https://github.com/Frix-x/klippain/blob/3f65109049c27deae0e089a8dfa9ec699feb90a5/user_templates/mcu_defaults/main/BTT_SKR_Mini_E3_v2.cfg) inside
- Added BTT Manta v1.0 and v1.1 MCU templates (a324734d2436a6b2dd4658d87706082a6d201484)
- Added a Klippain startup check system to validate specific incompatibilities in the user config and apply some initializations (like for the TMC2240) (9068790cb8e913b1346684bb11feb44eccad51c4 b6f36a535d59f681e4269d7036f3400383f9e67d ca4e3c9aeac05de76002ca700db75c6d81ff6865 6f4980f4dc1fdde542af0c5e6a6d589cbcf0d721 ad5d75efa410ec821f06a66593605f3155cb77ca)
- Improved the heatsoak phase of bed and chamber to avoid unecessary movements if they are already at temperature (d9f2a19001b0a44a626b172bf3f2d69556605816)
- Virtual probes types will now home directly at the center of the mesh (RRI position) if there is one computed. If it's not the case, it fallback to the center of the bed as it was previously (00d9d6e28bd70a5197a75900621989f706b525d1 d661c51a677cc8db44559265beca61b10a67468d)
- Automated the installation of the gcode_shell_command plugin during the Klippain installation. This plugin is mandatory for Klippain to work correctly and was previsouly installed manually using KIAUH (02a8be51c6b1dc8f1af64a0faa49046f53760a25 1a0acd4c0a2a7955ecdb256672404d2b721b7766)
- Added a variable to allow the customization of ooze time after a purge over the bucket (79a86b29bbccccd93bc5c966536e4dbf851a6456)
  > **Warning**:
  > 
  > **Breaking change**: you need to add the new variable to your variables.cfg: https://github.com/Frix-x/klippain/blob/79a86b29bbccccd93bc5c966536e4dbf851a6456/user_templates/variables.cfg#L199
- Added Beacon as a virtual probe option in Klippain. Now it's fully supported, you just need to add the plugin to Klipper,  enable the file in printer.cfg and add the proper [beacon] section to your overrides.cfg (2aab85516013ef688c9260b95ff22cc1dfff675f)

## Bugfixes and tweaks

- Raised the RPi temperature check to 95°C (instead of the previous 70°C that could lead to false "over-heating" errors) (fc75162b0688c9f6d25e3096f0c078e259249781)
- Fixed the START_PRINT actions override in the `override.cfg` user template. Now it should work correctly if you uncomment it (it was missing the macro header and 'gcode:' line to be able to work correctly). But be sure to get the last version in the last template (47487b4ee22fea5741bd9877594b4de21c6b8bd4)
- Commented the `[stepper_x]` default override that was injected automatically when running a toolhead MCU. This is to allow sensorless homing to work correctly (you can always uncomment the override if not using sensorless) (eb5083f62286c34aa8f71776a6229a1b04c3a304)
- Updated the Auto Z calibration configuration file to follow the config changes in v1.0.0 of the plugin (512a56d8c810f00c360e2e84bb8446b76aa2c747)
- Fixed the `plot_graph.sh` script to account for a bug introduced in Klippain v3: now if the resonance measuring position is added in the filename, it's removed by the script before processing to allow the correct behavior and correct naming of the file in the result folder (f1cf77a1f3381e749403b3c0a9b279db2d5db402)
- Fixed the race condition in the input shaper worflow that could happens on slow SD card, Pi clone or cheap alternatives mainly when running the belts test and resulted in csv files but no png images generated and an error in the console (e40edf2a26167bbbac07652c28c9ca312c61c51e)
- Fixed the input shaper workflow macro to allow the use of standard parameters. This also fix at the same time a rare bug where a wrong frequency per hz value were used after trying to use EXCITATE_AXIS_AT_FREQ (c7d8b2bfd7e107225ebbd0ae7e9a1b662a464fca fea91cdf8ee66d5331325a90e4f654dd0364d926)
- Fixed `position_endstop` invalid when using the combination of TAP on Trident (670551ac077bf20addffb831938a8d12a7e7e1de)
- Fixed the BTT SKR E3mini v2 MCU template that was partially broken (e160a2ca4376780acddcdf86199ee7875544d0ea)
- Updated a broken link in the adaptive bed mesh documentation (621bd936d1e3cea40e5a5bd8d7b994680e6ed2e5)
- Updated all the MCU templates to include directly the pin remapping as they are "static" overrides and it make sense to put them directly in the `mcu.cfg` file instead of `overrides.cfg` file. This have no impact on the functionnality but will help to avoid cluttering the `overrides.cfg` file (d704ac9555cc0716a092f5b2924d86ca44b0c074 efcce5550419a462b622f4da083e18530b9e2a57)
- Added a necessary user confirmation during the first install of Klippain to avoid "accidental" installation of the system and fixed a bug when the user would install Klippain without wiping first its old config that would let some file artefacts (now everything is correctly wiped automatically after the backup)(74654cc37aff1a0160f88a32bf5ae74ea2ba62c0)
- Fixed a bug happening when no heatbed were defined. Now it's ok to have a printer that don't have one (f20f395cdc747a80735daf7d0bcc7049f1374518)
- Deleted the default initial_angle used in servo files as it's not needed and could lead to potential problems (6979609c933d9b9f150160fc676df6d859e808cf)
- Added a G28 Z after the tilt calibration to get a new and correct Z origin. This was lost during the big change of Klippain to get the modular START_PRINT. This step is mandatory to avoid missing the brush or any other components if the machine is subject to high gantry sag on motor poweroff (cfb29b923c50dd31da1488db2e66b215f6648da8)
- Fixed the Mellow Fly ERCF MCU template (cc4ee87f05f291461ea1809ea1434483720581eb)
- G28 Z with XY already homed will now behave correctly and just home Z without forcing a full G28 as it was previously the case (a5bea10ad7dd5595ba6f2afbfc9301fb1cc70038)


config/hardware/axis/XY/TMC/xy_TMC2209.cfg
```diff
@@ -0,0 +1,19 @@
...
+run_current: 0.8
+sense_resistor: 0.110
+stealthchop_threshold: 0
+
+[tmc2209 stepper_y]
+uart_pin: Y_TMCUART
+interpolate: False
+run_current: 0.8
+sense_resistor: 0.110
+stealthchop_threshold: 0
```
Surion79, line 19: is here an empty line missing at the end?

config/hardware/axis/XY/TMC/xy_TMC2209_V0specific.cfg
```diff
@@ -1,12 +1,11 @@
-[stepper_x]
-rotation_distance: 40
-microsteps: 64
-full_steps_per_rotation: 200
+# XY TMC2209 definition
 
+# Including first the default XY TMC2209 file
+[include xy_TMC2209.cfg]
+
+# Then applying overrides for specific V0 parameters from LDO to reduce VFAs
```
Surion79, line 6: oh, the formbot v0 kits are not delivered with LDO. shouldn't be some central info that this settings are LDO specific?
I can't unoverride settings for a formbot kit

config/software/z_calibration.cfg
```diff
@@ -20,15 +20,10 @@ switch_xy_position: -1,-1
 # Switch offset (default D2F-5: 0.5mm and SSG-5H: 0.7mm)
 # Smaller number means higher nozzle to the PEI
 switch_offset: 0.50
-max_deviation: 0.750
+offset_margins: -0.750, 1.0
```
Surion79, line 23: does that work? `-0.750, 1.0`

docs/features/adaptive_bed_mesh.md
Surion79, line 1: should it be mentioned that this method is not working with ercf?

user_templates/overrides.cfg
```diff
@@ -24,26 +24,43 @@
...
+# control: pid
+# pid_kp: 26.015
+# pid_ki: 1.770
+# pid_kd: 95.603
+
+# [heater_bed]
+# control: pid
+# pid_kp: 38.771
+# pid_ki: 1.746
+# pid_kd: 215.177
```
Surion79, line 52: i recommend adding some info regarding how to obtain the values and that the save_config doesn't work as this section is not placed in printer.cfg (then save config would work)

Surion79 (commented):
some minor comments
some PRs (like the last SB2240 one) is missing

config/hardware/axis/XY/TMC/xy_TMC2209.cfg
Frix-x: No, it's here. Just github don't show it. FYI, when it miss an empty newline at the end, Github show a red symbol instead

config/hardware/axis/XY/TMC/xy_TMC2209_V0specific.cfg
Frix-x: Yes, using a standard V0 kit should use the standard V0 file. This specifics are used only for the problematic VFA generated by first generation of LDO motors (kit or not kit as you can buy them also by yourself). There were replaced and this should not be needed by newer generation of motors.

I can add a comment like : "if you use LDO motors on your V0 and get a lot of VFAs, try to use the V0 specific file instead"

config/software/z_calibration.cfg
Frix-x: Yes, in the last version of the plugin (>v1.0.x) the `max_deviation` is deprecated in favor of the newly added `offset_margins` parameter.

FYI, currently the Z calibration plugin is broken since 2 weeks in Klippain due to this change (solved by replacing the full file in overrides, so not a big deal but still)

user_templates/overrides.cfg
Frix-x: I indeed want to add a special comment somewhere to say that SAVE_CONFIG do not work with Klippain.

(It's easier to do that instead of adding PID to printer.cfg and/or some of the other things as then we would need to explain: this works using SAVE_CONFIG, but this does not, etc... and confuse people to "why". So I just prefer to keep the files clean and easy and say "SAVE_CONFIG" doesn't work in Klippain and that's it)

docs/features/adaptive_bed_mesh.md
Frix-x: Mhmm yes we can add this as a note indeed :)

install.sh
```diff
@@ -3,9 +3,11 @@
 ###### AUTOMATED INSTALL AND UPDATE SCRIPT ######
 #################################################
 # Written by yomgui1 & Frix_x
-# @version: 1.2
+# @version: 1.3
 
 # CHANGELOG:
+#   v1.3: added a warning on first install to be sure the user wants to install klippain and fixed a bug
+#         where some artefacts of the old user config where still present after the install (harmless bug but not clean)
```
Surion79, line 10: is the info regarding gcode shell missing?

Surion79 (approved):
just one remark: the gcode shell script installation isn't mentioned in the 1.3 changelog in the installation file.

install.sh
Frix-x: This is fixed in the last commit

Frix-x:
Ok, I think we are near the release of Klippain v4 now.
Can everybody involved here review and test to see if everything is alright?

elpopo-eng (changes requested):
With a high value for max_extrude_cross_section, I encountered an error in klipper (`stepper.error: Internal error in stepcompress`).
It's therefore unusable as it is with ercf because we need this large value.
As we discussed, it would be a good idea to use the slicer's maximum flow rate.

Frix-x:
> With a high value for max_extrude_cross_section, I encountered an error in klipper (`stepper.error: Internal error in stepcompress`). It's therefore unusable as it is with ercf because we need this large value. As we discussed, it would be a good idea to use the slicer's maximum flow rate.

Do you have more info in which macro this was a problem ?
FYI, I just pushed a new prime line system as it was probably one problematic point (it was not erroring but behaving very very very slowly almost not moving with a prime line of something like 2 minutes...). Now it should be more smart and should work for everyone :)

elpopo-eng:
> Do you have more info in which macro this was a problem ? FYI, I just pushed a new prime line system as it was probably one problematic point (it was not erroring but behaving very very very slowly almost not moving with a prime line of something like 2 minutes...). Now it should be more smart and should work for everyone :)

The error occured in `_MODULE_PRIMELINE` on the `M400`

Frix-x:
> The error occured in _MODULE_PRIMELINE on the M400

Ok so this should not be a problem anymore with the last commit :)

macros/base/start_print.cfg
```diff
@@ -163,14 +163,47 @@ gcode:
...
     {% set prime_line_length = printer["gcode_macro _USER_VARIABLES"].prime_line_length|float %}
+    {% set prime_line_purge_distance = printer["gcode_macro _USER_VARIABLES"].prime_line_purge_distance|float %}
     {% set prime_line_flowrate = printer["gcode_macro _USER_VARIABLES"].prime_line_flowrate|float %}
     {% set klippain_ercf_enabled = printer["gcode_macro _USER_VARIABLES"].klippain_ercf_enabled %}
-    {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
 
-    {% set max_extrude_cross_section = printer.configfile.settings.extruder.max_extrude_cross_section|float %}
-    {% set Sp = (prime_line_flowrate / max_extrude_cross_section) * 60 |float %}
+    {% set max_extrude_cross_section = printer["configfile"].config["extruder"]["max_extrude_cross_section"]|float %}
+    {% set filament_diameter = printer["configfile"].config["extruder"]["filament_diameter"]|lower %}
```
elpopo-eng, line 176: should be a float instead of a string (lower)

macros/base/start_print.cfg
```diff
@@ -393,3 +482,54 @@ gcode:
...
+    {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
+
+    {% if status_leds_enabled %}
+        STATUS_LEDS COLOR="HEATING"
+    {% endif %}
+    {% if verbose %}
+        RESPOND MSG="Pre-heating the nozzle to a safe temperature..."
+    {% endif %}
+
+    M109 S{safe_extruder_temp}
```
elpopo-eng, line 501: This is a problem with low inertia hotend.
As we ran the PID at high temperatures and the algorithm is not optimal, at low temperatures the measurement oscillates wildly and we get stuck here for a few minutes until the temperature luckily stabilizes.

Frix-x: Yes this hacky patch was done for this purpose :)

macros/base/start_print.cfg
Frix-x: Oops, my bad, I'll fix that!

Frix-x: C'est fait !

Frix-x:
Ok, now I have nothing else to add and you all have told me that it's working good for you!
Release planned for this afternoon :)

PR #190, opened 2023-05-09
SB2240 ADXL CS pin correction

jurikaur:
Pin naming correction not CS but should be NSS.


Surion79:
i was pretty sure that i made a PR yesterday for it. Found it too

PR #191, opened 2023-05-09
Fix BTT SKR Mini E3 v3 config

ndanyluk:
Hi! In #177 I realized that the config path I had included in my MCU default path was broken.  I fixed and verified on my printer, but didn't have a chance to commit until after the PR was merged.

This PR fixes that MCU template file.


Frix-x:
Opps, that's true. I forgot to check for squashed commit... Too late ^^
I'll also merge this correction. Thanks for your reactivity :)

PR #192, opened 2023-05-09
Correcting pin assignment.

jurikaur:
As BTT manual describes:
    FAN1 - PA0 - Hotend FAN
    FaN2 - PA1 - Part cooling FAN
If somebody follows BTT pdf Guide then it is good to use same assignment.


Frix-x:
Thanks!
Do you see other things to fix for the SB2240 before the release of Klippain v4?

jurikaur:
Right now functionality seems ok. Managed to clog my nozzle and cannot test printing. Have to soak hotend in acetone. Test print has to wait. 
Tested fans, leds, extruder, hotend heater and thermistor. Not tested adxl, long print.

PR #195, opened 2023-05-09
Tmc2240 in sb2240cfg

Surion79:
i commented the macro and wrote a message to override the _init_tmc2240 if settings must be done. 
i moved the "driver*" settings of the tmc2240 section from the "protected" section to the mcu section, so if specific "driver*" configs are no longer needed, they can be deleted there individually based on recommendations of e.g. Esoterical


config/hardware/extruder/TMC/e_TMC2240.cfg
```diff
@@ -3,32 +3,12 @@
...
 [tmc2240 extruder]
 cs_pin: E_CS
-spi_speed: 500000
 spi_software_sclk_pin: E_SCLK
 spi_software_mosi_pin: E_MOSI
 spi_software_miso_pin: E_MISO
-interpolate: False
 run_current: 0.6
-rref: 12000
 stealthchop_threshold: 99999
```
Frix-x, line 12: Can you move this line also? I feel that stealchop should not be enablde by default

macros/miscs/startup.cfg
```diff
@@ -12,34 +12,35 @@ gcode:
...
-    SET_TMC_FIELD STEPPER=extruder FIELD=OVERTEMPPREWARNING_VTH VALUE=2885 # 7.7 * 100 C + 2038
-    # The following is absolutely critical: set the overvoltage snubber to a sensible voltage.
-    # This should be set to about 0.8 V above your power supply's idle voltage.
-    # Your PSU voltage can be read from the TMC 2240 by issuing a GCODE command:
-    # DUMP_TMC stepper=extruder register=ADC_VSUPPLY_AIN
-    # The voltage is the value of adc_vsupply multiplied by 0.009732
-    # It is also possible to use adc_vsupply + 82 here, which works out to be the same.
-    {% set v = (24.7/0.009732)|int %}
-    SET_TMC_FIELD STEPPER=extruder FIELD=OVERVOLTAGE_VTH VALUE={ v }
+    RESPOND MSG="You have defined to have a TMC2240 installed. If you need specific gcode settings, please override macro '_INIT_TMC2240'"
```
Frix-x, line 17: This should not use RESPOND as it will spam user about that on each restart. Maybe just a comment is enough to say that these defaults fixes things based on Esoterical github but are not necessary for everyone and could be tweaked

Surion79:
I'll do it when i am home in 1 hr

Frix-x (approved):

Frix-x (approved):

PR #196, opened 2023-05-10
Add Preliminary Switchwire Support

ndanyluk:
This PR introduces preliminary support for the Voron Switchwire printer, bringing it in line with the other Voron printers as supported by Klippain.

Completed Changes:
[x] Move [printer] kinematics out of machine.cfg into new 'config/hardware/kinematics' config section (allows for CoreXY, CoreXZ and possibly Cartesian motion systems) (BREAKING CHANGE)
[x] Adds new axis types 'XZ' and 'Y' for Switchwire support without modifying existing files for CoreXY, besides printer.cfg
[x] Adds Voron Switchwire default Prusa Mk52 bed size
[x] Introduces z_virtual_endstop probe sub-types for Inductive and Dockable probes to support VSW stock config.

Changes Requiring Help:
[ ] The existing dockable probe macros don't seem to work in testing.  The printer attempts to probe Z before attaching the dock (which fails since the endstop will be triggered without the probe attached)

Functions Tested:
[x] Stepper buzz for all axis (PASS)
[x] XY motion and homing (PASS)
[ ] Z homing (FAIL, see above)
[ ] Full motion tests
[ ] Print_Start
[ ] Print_Pause
[ ] Print_End

Let me know what you think!  There's still a lot to do, but I thought this might be good to include as part of Klippain v4 if we can work out the probe issues quickly.


ndanyluk:
Hi @Frix-x, finally got to work on that SW PR!

One issue I haven't figured out yet that I could use a hand on is the homing macro.  I'm running with Klicky and it seems like the dockable probe macros attempt to home Z before deploying the dock (like if we have a normal Z end stop + probe).

Not sure how to make that work without breaking something

Surion79:
> Hi @Frix-x, finally got to work on that SW PR!
> 
> One issue I haven't figured out yet that I could use a hand on is the homing macro. I'm running with Klicky and it seems like the dockable probe macros attempt to home Z before deploying the dock (like if we have a normal Z end stop + probe).
> 
> Not sure how to make that work without breaking something

what probe did you uncomment in printer.cfg?

ndanyluk:
> > Hi @Frix-x, finally got to work on that SW PR!
> > One issue I haven't figured out yet that I could use a hand on is the homing macro. I'm running with Klicky and it seems like the dockable probe macros attempt to home Z before deploying the dock (like if we have a normal Z end stop + probe).
> > Not sure how to make that work without breaking something
> 
> what probe did you uncomment in printer.cfg?

A new dockable_virtual_endstop probe.  Essentially the dockable probe with stepper_z endstop pin set to probe:z_virtual_endstop like Voron Tap.

It's a new probe type for this PR but intended to have similar functionality to the existing dockable probe.  I also added an inductive_virtual_endstop.cfg for the case of the stock Switchwire inductive probe as well

config/hardware/axis/voron_switchwire_mk52.cfg
```diff
@@ -0,0 +1,18 @@
...
+position_max: 250
+
+[stepper_y]
+position_min: 0
+position_endstop: 240
+position_max: 240
+
+[stepper_z]
+position_max: 235
+position_min: -3.0
```
Surion79, line 18: would it be possible to rename it a bit more generic with just adding the bedsize? like the "default_<ZZZ>mm.cfg" or "default_<XXX>mmX<YYY>mm.cfg"

Surion79:
> > > Hi @Frix-x, finally got to work on that SW PR!
> > > One issue I haven't figured out yet that I could use a hand on is the homing macro. I'm running with Klicky and it seems like the dockable probe macros attempt to home Z before deploying the dock (like if we have a normal Z end stop + probe).
> > > Not sure how to make that work without breaking something
> > 
> > 
> > what probe did you uncomment in printer.cfg?
> 
> A new dockable_virtual_endstop probe. Essentially the dockable probe with stepper_z endstop pin set to probe:z_virtual_endstop like Voron Tap.
> 
> It's a new probe type for this PR but intended to have similar functionality to the existing dockable probe. I also added an inductive_virtual_endstop.cfg for the case of the stock Switchwire inductive probe as well

it sounds a bit like this is already available in klippain. but i just working through your PR the first time :)

config/hardware/axis/voron_switchwire_mk52.cfg
ndanyluk: We could do that.  I think that the big issue that needs to be solved is that the travel for the SW is from -10,240 on Y and the silkscreen print on the Prusa Mk52 bed is shown as 250 x & 210 y (there are a pair of locating pegs that restrict the safe y print area at 0 < Z < 10 to this value).

So do we put 250x240 which is the physical size of the bed? Or 250x210 which is the safe print area?

ndanyluk:
> > > > Hi @Frix-x, finally got to work on that SW PR!
> > > > One issue I haven't figured out yet that I could use a hand on is the homing macro. I'm running with Klicky and it seems like the dockable probe macros attempt to home Z before deploying the dock (like if we have a normal Z end stop + probe).
> > > > Not sure how to make that work without breaking something
> > > 
> > > 
> > > what probe did you uncomment in printer.cfg?
> > 
> > 
> > A new dockable_virtual_endstop probe. Essentially the dockable probe with stepper_z endstop pin set to probe:z_virtual_endstop like Voron Tap.
> > It's a new probe type for this PR but intended to have similar functionality to the existing dockable probe. I also added an inductive_virtual_endstop.cfg for the case of the stock Switchwire inductive probe as well
> 
> it sounds a bit like this is already available in klippain. but i just working through your PR the first time :)

Haha I'm really hoping it is! The reason I added two probe types was that I hit an error trying to simply use dockable.cfg.  I set an override for stepper_z endstop_pin to probe:z_virtual_endstop like the stock SW klipper config.  However, Klipper complained that z_endstop_pin was then invalid (set in the dockable/inductive configs).  

So my solution was to follow the Voron Tap cfg style and replace "endstop_position=0" with "endstop_pin=probe:z_virtual_endstop".  The config can run now without a klipper errors until Z homing

ndanyluk:
@Surion79 also thanks for taking a look through :)

Surion79:
I ddon't like to multiply the cfgs by config settings like virtual_endstop-version of prove xyz. I think there should be some sort of generic configuration like having that stepper_z config in a different cfg and reference it. But thats another decision for @Frix-x

ndanyluk:
Here's my variables.cfg for reference:
```
[gcode_macro _USER_VARIABLES]
# Enable verbose output to let the macro have a chat
variable_verbose: True

###################################################
# Feedrate and accels setpoints for all the macros
###################################################

## SPEEDS
variable_homing_travel_speed: 350
variable_travel_speed: 350
variable_z_drop_speed: 15
variable_brush_clean_speed: 100
variable_probe_dock_speed: 50

## ACCELS
## Those are specific accel settings for some of the macros
variable_homing_travel_accel: 3000
variable_tilting_travel_accel: 3000
variable_brush_clean_accel: 1500
variable_probe_dock_accel: 2000

## Accel to decel scale factor (default is 75% of the accel)
variable_accel_to_decel_factor: 0.75


#################################################
# Homing, start_print and end_print variables
#################################################

## Physical Z endstop pin position. This variable is used only if
## you don't use the automatic z_calibration plugin: if this plugin is
## installed, the z endstop pin position is directly extracted from the
## plugin config file (see config/software/z_calibration.cfg) or its user override.
variable_zendstop_position: 0, 0 # change it ONLY if not using the auto z calibration plugin !!!

## Force always a full homing and QGL/Z_TILT during the START_PRINT macro
variable_force_homing_in_start_print: False

## Z hop before homing to avoid grinding
## the bed due to the gantry sag
variable_homing_zhop: 5

## XY axis homing order and backoff distance after touching the endstops
variable_homing_first: "X" # can be set to "Y" first
variable_homing_backoff_distance_xy: -10, -10 # move 5mm in negative direction on both axis

## Percentage of run_current used while sensorless homing (if used)
variable_sensorless_current_factor: 75

## margin to avoid the probe dock when homing
variable_probe_dock_margin_xy: 0, 0

## Temperature of the extruder used to preheat the chamber temperature and do some START_PRINT actions
variable_safe_extruder_temp: 170

## Prime line parameters
variable_prime_line_xy: 2.5, 5
variable_prime_line_length: 40
variable_prime_line_flowrate: 5 # mm3/s used for the prime line

## Park position used when pause, end_print, etc...
variable_park_position_xy: 5, 235
variable_park_lift_z: 50

## Automatically disable motors in the END_PRINT macro
variable_disable_motors_in_end_print: False

## Automatically turn-off heaters in the END_PRINT macro
variable_turn_off_heaters_in_end_print: True


#########################################################
# Dockable probe variables (if available in the machine)
#########################################################

## Bed size used to avoids doing a probe_accuracy or probe_calibrate outside the bed
## This has no effect in the macros beside a safety check (not mandatory to be changed)
variable_min_bed_xy: 0, 0
variable_max_bed_xy: 250, 240

## Minimum safe Z height to attach/detach probe
variable_probe_min_z_travel: 25

## Position of the probe dock
variable_probe_dock_location_xy: 207.5, 120

## Servo angles used to define the retracted and deployed positions
## of the dock (if applicable). These variables are only used if a probe servo
## is included in the config and have no effect if it's not the case
variable_probe_servo_angle_retracted: 92
variable_probe_servo_angle_deployed: 35

## Positions of the toolhead when docking/undocking the probe
## See diagram below for help
variable_probe_before_attach_position: "front"
variable_probe_after_attach_position : "front"
variable_probe_before_dock_position : "front" # generaly same as probe_after_attach_position
variable_probe_after_dock_position : "left"

variable_probe_move_attach_length: 0
variable_probe_move_dock_length: 40

##    Y
##    ^ 
##    |          back
##    |           ^
##    |   left  < O >  right
##    |           v
##    |         front
##    |_ _ _ _ _ _ _ _ _ _ _ _> X

## If there is an error during a tilting procedure (QGL or Z_TILT_ADJUST) or during a BED_MESH_CALIBRATE
## then automatically dock the probe before stopping all actions. This avoid letting the probe
## very close to a very hot bed doing nothing as this could destroy the probe microswitch
variable_autodock_on_probe_error: True


##########################################################
# Voron TAP probe variables (if available in the machine)
##########################################################

variable_tap_max_probing_temp: 150
variable_tap_deactivation_zhop: 5 # this is used to Z hop before restoring the temperature to avoid burnt PEI


##########################################
# Material and specific print parameters
##########################################

## default START_PRINT parameters
variable_print_default_bed_temp: 105
variable_print_default_extruder_temp: 240
variable_print_default_chamber_temp: 0
variable_print_default_chamber_max_heating_time: 15
variable_print_default_soak: 8
variable_print_default_material: "XXX"

## Material configuration parameters applied during START_PRINT by using the slicer MATERIAL variable
## FYI, retract paramaters are used only if firmware retraction is enabled, filter speed (in %) is used if
## there is a filter installed on the machine, etc...
## If you are using another material, just extend the list bellow with a new material and everything should work :)
variable_material_parameters: {
        'PLA': {
            'pressure_advance': 0.0525,
            'retract_length': 0.75,
            'unretract_extra_length': 0,
            'retract_speed': 40,
            'unretract_speed': 30,
            'filter_speed': 0,
            'additional_z_offset': 0
        },
        'PET': {
            'pressure_advance': 0.0650,
            'retract_length': 1.4,
            'unretract_extra_length': 0,
            'retract_speed': 30,
            'unretract_speed': 20,
            'filter_speed': 0,
            'additional_z_offset': 0.020
        },
        'ABS': {
            'pressure_advance': 0.0480,
            'retract_length': 0.5,
            'unretract_extra_length': 0,
            'retract_speed': 40,
            'unretract_speed': 30,
            'filter_speed': 80,
            'additional_z_offset': 0
        }
    }


# Other hardware options used in the macros

## Brush and purge bucket available in the machine
## If this is not available in the machine, set both positions
## variables to "-1, -1, -1" for safety
variable_purge_and_brush_enabled: False
variable_force_homing_before_brush: False # Add an homing of the Z axis before a clean to be sure to not miss the brush
variable_brush_over_y_axis: True # Cleanup is done moving the extruder along the Y axis, then X axis. If false, only X axis is done
variable_brush_xyz: -1, -1, -1 # Position of the brush center for nozzle cleaning
variable_purge_bucket_xyz: -1, -1, -1 # Purge bucket position
variable_purge_distance: 30 # Amount to purge

## Servo angles used to define the retracted and deployed positions
## of the purge bucket and brush (if applicable). These variables are only used if a purge
## bucket servo is included in the config and have no effect if it's not the case
variable_purgeclean_servo_angle_retracted: 0
variable_purgeclean_servo_angle_deployed: 90


## White light parameters (if installed in the machine)
variable_light_intensity_start_print: 100
variable_light_intensity_printing: 30
variable_light_intensity_end_print: 0


## Patch the M190/M109 commands to avoid some wait time while the temperature
## settle on very low thermal latency devices (such as the BambuLabs hotend)
variable_fix_heaters_temperature_settle: False


## Resonnance testing position (if an accelerometer is installed in the machine)
## If test point is let to -1,-1, it default to the center of the bed at 50mm high
variable_resonnance_test_point_xy: -1, -1
variable_resonnance_test_z_clearance: 50


## SteathBurner minidisplay and case leds colors (if installed in the machine)
variable_status_leds_colors: {
        'logo': {
            'busy': {'r': 0.4, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'cleaning': {'r': 0.0, 'g': 0.02, 'b': 0.5, 'w': 0.0},
            'calibrating_z': {'r': 0.8, 'g': 0., 'b': 0.35, 'w': 0.0},
            'heating': {'r': 0.3, 'g': 0.18, 'b': 0.0, 'w': 0.0},
            'homing': {'r': 0.0, 'g': 0.6, 'b': 0.2, 'w': 0.0},
            'leveling': {'r': 0.5, 'g': 0.1, 'b': 0.4, 'w': 0.0},
            'meshing': {'r': 0.2, 'g': 1.0, 'b': 0.0, 'w': 0.0},
            'on': {'r': 0.8, 'g': 0.8, 'b': 0.8, 'w':1.0},										  
            'off': {'r': 0.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'printing': {'r': 1.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'standby': {'r': 0.01, 'g': 0.01, 'b': 0.01, 'w': 0.1},
            'error': {'r': 0.6, 'g': 0.0, 'b': 0.0, 'w':0.0},
        },
        'nozzle': {
            'heating': {'r': 0.8, 'g': 0.35, 'b': 0.0, 'w':0.0},
            'off': {'r': 0.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'on': {'r': 0.8, 'g': 0.8, 'b': 0.8, 'w':1.0},
            'standby': {'r': 0.6, 'g': 0.0, 'b': 0.0, 'w':0.0},
            'error': {'r': 0.6, 'g': 0.0, 'b': 0.0, 'w':0.0},
        },
        'caselight': {
            'busy': {'r': 0.4, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'cleaning': {'r': 0.0, 'g': 0.02, 'b': 0.5, 'w': 0.0},
            'calibrating_z': {'r': 0.8, 'g': 0., 'b': 0.35, 'w': 0.0},
            'heating': {'r': 0.3, 'g': 0.18, 'b': 0.0, 'w': 0.0},
            'homing': {'r': 0.0, 'g': 0.6, 'b': 0.2, 'w': 0.0},
            'leveling': {'r': 0.5, 'g': 0.1, 'b': 0.4, 'w': 0.0},
            'meshing': {'r': 0.2, 'g': 1.0, 'b': 0.0, 'w': 0.0},
            'on': {'r': 0.8, 'g': 0.8, 'b': 0.8, 'w':1.0},
            'off': {'r': 0.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'printing': {'r': 1.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'standby': {'r': 0.01, 'g': 0.01, 'b': 0.01, 'w': 0.1},
            'error': {'r': 0.6, 'g': 0.0, 'b': 0.0, 'w':0.0},
        },
        'minidisplay': {
            'on': {'r': 0.0, 'g': 0.2, 'b': 0.4, 'w':1.0},
            'off': {'r': 0.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'error': {'r': 0.4, 'g': 0.0, 'b': 0.0, 'w':0.0},
        },         
        'thermal': {
            'hot': {'r': 1.0, 'g': 0.0, 'b': 0.0, 'w': 0.0},
            'cold': {'r': 0.3, 'g': 0.0, 'b': 0.3, 'w': 0.0}
        }
    }


## Do not remove the next lines
gcode:

[save_variables]
filename: ~/printer_data/config/save_variables.cfg
```

Surion79:
can you post the klipper log with your full config? the includes add variables to it, so i only see your variables.cfg but not all variables added by including files

Surion79:
in post v4 release, there might be a topic to include some additional debug level so such issues are better traceable.
you can spam the specific sections with ` { action_respond_info("message") }` to follow the flow. but if you like a teamwork, i can offer you some "live" cooperation on this one via teams or other tools

ndanyluk:
> in post v4 release, there might be a topic to include some additional debug level so such issues are better traceable. you can spam the specific sections with ` { action_respond_info("message") }` to follow the flow. but if you like a teamwork, i can offer you some "live" cooperation on this one via teams or other tools. But it would be debugging with the action_respond_info to find the missing config value or setting. might be boring :D

Unfortunately I got stuck late at work today but I appreciate it haha :)

I think I found the problem though.  It's a two part problem:

1. After identifying that with Z probed (manual probe attach) `_attach_probe` works, I realized that the stock homing_override doesn't actually include a case where the dockable probe is the virtual endstop:

From homing_override.cfg
```
 {% if probe_type_enabled == "vorontap" %}
            ACTIVATE_PROBE
        {% endif %}
        G28 Z0
        {% if probe_type_enabled == "vorontap" %} # And restore the original temperature
            DEACTIVATE_PROBE
        {% endif %}
```

I think adding a new check for dockable probe & virtual endstop would solve this.

2. The dockable _attach_probe macro requires Z to be homed before grabbing the probe (as is the case with Klicky + physical endstop).  The source Klicky macros get around this by doing a Z hop whether Z is homed or not just to be safe.

I have an idea for how to fix but it's been a long day and I might not get to it until tomorrow unfortunately

Surion79:
> 

@ndanyluk 
Regarding 1. There are already automated checks for virtual z endstop. have you seen them?
Regarding 2. Frix-x and I were discussing some z hop topics in a different PR. so IMO it would be better to discuss how to solve your issue in general instead of working something different of the current functionality.

ndanyluk:
> Hello,
> 
> So first of all thanks for this PR ! It's a very great work :) As a first step before talking about names and things to finalize the PR, I would go for solving the problem for the Z and things. Then we will be able to refactor.
> 
> So here is how I would do it: currently there is a "probe_type" defined as a user variable in each probe config file. This variable is then used as a way to know which action to do in every homing and probing macros. So to solve the problem:
> 
> 1. You will need to differentiate the "dockable probe" and "dockable probe virtual": add this "virtual" keyword to make new probes types in your new files.
> 2. Then look in the macros for all the `{% if probe_type_enabled == "vorontap" %}` (CTRL+F in all the files) and duplicate this kind of test to add some conditions for the new virtual probes types. I suggest you to base your new conditions to the voron TAP as it's already a virtual probe working correctly.
> 
> This should define a proper behavior for the new probes types and allow you to home Z as well :)

Thanks!! Really love the project and happy to contribute :)

I've added a few commits that I think will properly address the Z issues (I'm at work so can't test until tonight when I get home).  Overall, the new config does the following:

* Adds a new probe .cfg option for printer.cfg called `dockable_virtual_endstop`
* That config inherits properties from the normal dockable probe, but sets the `probe_type_enabled = dockable_virtual_endstop` 
* Added a new dockable_virtual_endstop macros cfg to replace the stock _attach_probe and _dock_probe functions that avoids the requirement for Z to be homed (since it can't be in this configuration - modified the code slightly to match the way Klicky does it)
* Added a check to the homing_override that looks for `probe_type_enabled == dockable_virtual_endstop` and activates the probe

Hoping this does the trick, but I'll report back tonight!

Surion79:
> > Hello,
> > So first of all thanks for this PR ! It's a very great work :) As a first step before talking about names and things to finalize the PR, I would go for solving the problem for the Z and things. Then we will be able to refactor.
> > So here is how I would do it: currently there is a "probe_type" defined as a user variable in each probe config file. This variable is then used as a way to know which action to do in every homing and probing macros. So to solve the problem:
> > 
> > 1. You will need to differentiate the "dockable probe" and "dockable probe virtual": add this "virtual" keyword to make new probes types in your new files.
> > 2. Then look in the macros for all the `{% if probe_type_enabled == "vorontap" %}` (CTRL+F in all the files) and duplicate this kind of test to add some conditions for the new virtual probes types. I suggest you to base your new conditions to the voron TAP as it's already a virtual probe working correctly.
> > 
> > This should define a proper behavior for the new probes types and allow you to home Z as well :)
> 
> Thanks!! Really love the project and happy to contribute :)
> 
> I've added a few commits that I think will properly address the Z issues (I'm at work so can't test until tonight when I get home). Overall, the new config does the following:
> 
> * Adds a new probe .cfg option for printer.cfg called `dockable_virtual_endstop`
> * That config inherits properties from the normal dockable probe, but sets the `probe_type_enabled = dockable_virtual_endstop`
> * Added a new dockable_virtual_endstop macros cfg to replace the stock _attach_probe and _dock_probe functions that avoids the requirement for Z to be homed (since it can't be in this configuration) based on the way that the Klicky probe does it)
> * Added a check to the homing_override that looks for `probe_type_enabled == dockable_virtual_endstop` and activates the probe, then expanded the check after homing to deactivate the probe if either Voron Tap or dockable_virtual_endstop is in use
> 
> Hoping this does the trick, but I'll report back tonight!

could you please edit your initi post? or is it still the current issue? I really like your work on this. You seem very motivated :)

Frix-x:
Wow, this is very good news! I've delayed the release of the v4 branch to include this inside at the same time if it's ok for you? So "official CoreXZ and Cartesian printer support, compatible with the Switchwire" will be part of the v4 :)

Could you also add at the same time a virtual probe type for the inductive probe? This would help cover the "standard" and stock switchwire as per Voron BOM. Also I think it could help using Klippain on basic and standard Cartesian kinematics printers such as ender5 and co :)

Frix-x:
> Haha! It works!! There's an odd behavior (to me) where my Klicky dock deploys automatically at printer startup, but otherwise the modifications work! Updating original post with testing status now 😀

I think this servo dock problem is probably due to the init position of the servo in the configuration. You can override it if needed. FYI, these are the defaults included in the `probe_servo.cfg` file regarding this:
```
[servo probeservo]
pin: SERVO_PIN
initial_angle: 0
maximum_servo_angle: 90
minimum_pulse_width: 0.0011
maximum_pulse_width: 0.0021
```

Frix-x:
And finally, if ecerything is checked and run properly, can you squash your commits to transform them in only one commit ? This will be more clean and will allow me to work on it for the proper integration before releasing Klippain v4 :)

Frix-x:
No problem, this is a hobby project, take the time you need :) Cheap servos are always a bit finicky and I already broke a couple of them... Same symptoms: overheating and not working correctly anymore (like random movements length, etc...)

On my side I'll also try some regression tests on both my V2.4 and Trident machines.

ndanyluk:
Thanks :)

Regarding the servos, I added a commit to remove the initial_angle line from the config.  Klipper does not require it and default behavior is to not send an angle at startup.  Since most probes/purge buckets require locking the arm into a specific set of positions (i.e. klicky dock or purge bucket arm) and we define those in user_variables.cfg, I think the line might not be needed anymore.

Since it's set to 0 degrees by default, it pushes the servo to one extreme as well since many use 0 to 180 degrees as their range.

Surion79:
> Thanks :)
> 
> Regarding the servos, I added a commit to remove the initial_angle line from the config. Klipper does not require it and default behavior is to not send an angle at startup. Since most probes/purge buckets require locking the arm into a specific set of positions (i.e. klicky dock or purge bucket arm) and we define those in user_variables.cfg, I think the line might not be needed anymore.
> 
> Since it's set to 0 degrees by default, it pushes the servo to one extreme as well since many use 0 to 180 degrees as their range.

very good thinking, Klippain should not do bad stuff by default, damaging components. :)

Frix-x:
Hi, @ndanyluk, do you have any news on this topic? :)
I've tested part of this branch with good success on my machine and waiting for you to tell me when you're ready :)

ndanyluk:
> 

Hi Frix, unfortunately nothing yet.  It's been a crazy week at work for me and not sure I'll be able to adequately test the code until after next week.  I'll try to test this weekend if I can!

Frix-x:
This was merged in the branch https://github.com/Frix-x/klippain/tree/multi-kinematics
I'll continue the work in this new branch :)

PR #200, opened 2023-05-10
Ercf rework - draft for review and discussion

Surion79:
This draft PR for discussing changes in the ercf rework linked to code.
i PRed it to develop, so only the relevant commits are shown here


Surion79:
@Benoitone I haven't found anything in HH regarding the function `_DO_ERROR_ERCF_PAUSE`
I thought this is required by klippain so i copied the last version of it from the PR to my printer. but if load filament fails, the ercf is stuck in lock state and print is cancelled. I don't recognize this behavior using current HH without klippain.

Benoitone:
> @Benoitone I haven't found anything in HH regarding the function `_DO_ERROR_ERCF_PAUSE` I thought this is required by klippain so i copied the last version of it from the PR to my printer. but if load filament fails, the ercf is stuck in lock state and print is cancelled. I don't recognize this behavior using current HH without klippain.

I had to create this action. Indeed, the error when loading a filament puts the print on hold. But does not stop the execution of the current macro. Suddenly all the print start is executed as if there was no problem and the pause is done only after... So I think it's useful to keep this function.
I also saw that you commented on the servo up at the start. It has already happened to me that the servo is in the down position when printing is started and this can be a problem. Normally it's more or less integrated in HH but I'm not 100% sure

Surion79:
Only thing missing is the discussion about the initial anchorpoint for ercf

Benoitone:
Can I push one commit and if you are not agree go back? I'm not pro with github...

macros/base/start_print.cfg
```diff
@@ -349,9 +349,11 @@ gcode:
...
         {% if ercf_reset_stats_on_start_print %}
             ERCF_RESET_STATS
-        {% endif %}            
-        ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
-        ### _DO_ERROR_ERCF_PAUSE
+        {% endif %}
+        {% if printer.ercf.tool|int != INITIAL_TOOL or printer.ercf.filament != "Loaded" %}
+            ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
+        {% endif %}
+        _DO_ERROR_ERCF_PAUSE
```
Surion79, line 356: where did you readd this function?

macros/base/start_print.cfg
```diff
@@ -349,9 +349,11 @@ gcode:
     {% if ercf_enabled %}
         {% if ercf_reset_stats_on_start_print %}
             ERCF_RESET_STATS
-        {% endif %}            
-        ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
-        ### _DO_ERROR_ERCF_PAUSE
+        {% endif %}
+        {% if printer.ercf.tool|int != INITIAL_TOOL or printer.ercf.filament != "Loaded" %}
```
Surion79, line 353: This is done automatically by HH. why would you want to repeat its internal function?

user_templates/variables.cfg
```diff
@@ -176,8 +176,8 @@ variable_material_parameters: {
 ################################################
 ## this section is only considered if an ercf is installed and configured
 
-variable_ercf_unload_on_cancel_print: True
-variable_ercf_unload_on_end_print: True
+variable_ercf_unload_on_cancel_print: False
+variable_ercf_unload_on_end_print: False
```
Surion79, line 180: not sure if you want to invert the ercf functionality as it is now in klippain. The defaults of the config is the current behavior.

Surion79:
somehow this is getting lost. i asked for discussion for a solution :D instead just delete stuff and don't comment or solve it a different way.

macros/base/end_print.cfg
```diff
@@ -7,16 +7,19 @@ gcode:
...
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
 
     PARK
 
-    {% if ercf_enabled and ercf_unload_on_end_print %}
-       # unload filament and park into ercf
-        ERCF_EJECT
+    {% if ercf_enabled %}
+        {% if unload|int == 1 or ercf_unload_on_end_print %}    # leaves the choice between slicer end print gcode or Klippain variables
```
Surion79, line 19: what takes precedence? slicer or config? with slicer i want to not unload the filament but the variable is defined for it.

Benoitone:
> somehow this is getting lost. i asked for discussion for a solution :D instead just delete stuff and don't comment or solve it a different way.

I'm so sorry...

macros/base/end_print.cfg
Benoitone: I thought that in this way it would allow the user either to leave more or less permanently the ejection at the end of the print (with your variable) or to allow it to be defined according to the profile of the printer in the slicer by putting `END_PRINT UNLOAD_AT_END=1` in the End G-code of the slicer

user_templates/variables.cfg
Frix-x: In this case I think we should use the slicer entry primarly. If this is not available (in case the use did not put it the custom start g_code), then fall back to use the variable instead

macros/base/start_print.cfg
Frix-x: This function is a custom Klippain function added to be able to stop the START_PRINT routine in case there is an error on the ERCF

macros/base/end_print.cfg
Frix-x: As stated in another comment. I think it's needed to define a clear order and here it should be slicer priority: if END_PRINT `UNLOAD_AT_END` is defined, use it.
If it's not the case, then fallback to the variable.

Also please rename the variable to ERCF_UNLOAD_AT_END (to not confuse the users about it)

config/hardware/ercf.cfg
```diff
@@ -4,13 +4,6 @@
 variable_ercf_enabled: True
 gcode:
 
-# Include all files which are created by the install script of Happy Hare
-[include ../../ercf_hardware.cfg]
-[include ../../ercf_menu.cfg]
-[include ../../ercf_software.cfg]
-[include ../../ercf_parameters.cfg]
-[include ../../ercf_vars.cfg] ## TODO: check if this is better placed centrally in klippain and not introduced only via ERCF in klippain
```
Frix-x: If you remove that from here. How do you include the HH files?

Benoitone: I include them in printer.cfg. because of the symlink it's not possible to make otherwise :(
unless you have a genius idea.

Surion79: We could put an ercf.cfg in the mainfolder so we have only one include in printer.cfg

Benoitone: It's another solution

macros/base/end_print.cfg
Benoitone: I push an update... I can't test now. Are you OK?

user_templates/variables.cfg
Benoitone: I can give it back to True if you prefer?

config/hardware/ercf.cfg
Benoitone: I know it's not very clean but I've never found a solution to this problem with the symlink

Surion79: how do we handle klipperscreen happy hare edition? :D if we are already talking about optionals :P

Surion79:
@Frix-x Could we remove the Material settings if ercf is enabled because of multi filament?

Surion79:
@Benoitone could you check the updated end print? I think your change gave me an idea for the "perfect" solution: Only if the parameter is not provided, the variable in klippain is used

Surion79:
@Benoitone can you please elaborate regarding the error function which was implemented in klippain but not in ercf? 
When i had issues with ercf, after primeline (well wasn't possible because missing filament) it paused because of error while loading.
The do error function is throwing three errors at once. so it is imho not clean. 
Should we have a function that errors before prime lining and pauses the print? the current solution is not clean IMO

Benoitone:
> @Benoitone could you check the updated end print? I think your change gave me an idea for the "perfect" solution: Only if the parameter is not provided, the variable in klippain is used

In this case, can we change the default value to 1 instead of 99?

Surion79:
> > @Benoitone could you check the updated end print? I think your change gave me an idea for the "perfect" solution: Only if the parameter is not provided, the variable in klippain is used
> 
> In this case, can we change the default value to 1 instead of 99?

no, because then you can't differ between an explicit set of the variable by parameter or not provided. so putting it on neither 0 nor 1 makes it perfect

Benoitone:
> @Benoitone can you please elaborate regarding the error function which was implemented in klippain but not in ercf? 
> 
> When i had issues with ercf, after primeline (well wasn't possible because missing filament) it paused because of error while loading.
> 
> The do error function is throwing three errors at once. so it is imho not clean. 
> 
> Should we have a function that errors before prime lining and pauses the print? the current solution is not clean IMO

If you have a cleaner solution to prevent the start_print from continuing right after the failed load? I had no other way to avoid having to wait until the end and avoid all the next steps before it pauses after the primeline at the end of the start_print sequence

config/hardware/ercf.cfg
Benoitone: I don't need to add anything to klippain for install HH KlipperScreen (same for crowsnest...)

user_templates/variables.cfg
Frix-x: No, just do it in two steps :

```
{ if param of slicer exist }
  USE PARAM OF SLICER
{ else }
  FALLBACK TO THE USE OF THE VARIABLE
```

Frix-x:
> What about a M400 and then check if printer.ercf.filament == "Loaded" and if not loaded then pause? Or even add a retry?

No it will not work. I think you miss the understanding of how template expansion is working (the Jinja).
So just a little bit of explanation: when you run a macro there is a two step process happening in Klipper:
  1. Klipper read the Jinja to create the gcode internally. This means that all the variables used in the Jinja are read at this specific instant. If you have for example a "for" loop to move axis, then it will just create the corresponding Gcode and replace it directly. All the "loops" are created at this moment
  2. Then when the template expansion is finished, there will be only standard gcode with all the values replaced, etc... This gcode will be executed with no "intelligence", just line by line. So no "specific checks" can be done at this moment.

At the time we wrote the ERCF error macro, the `START_PRINT` was just a monolitic big macro. So at START_PRINT call, during the template expansion phase the ERCF was ok and all the gocde generated properly. Then during the START_PRINT procedure, if for some reason the ERCF was erroring, as the template expansion was already done, there was nothing happening as the variable was not even read. The solution for that is to call another macro to do the check as this new macro will have its own context.

It's also for the same reason that `HEATSOAK_CHAMBER` and `_WAIT_CHAMBER_TEMP` is done in two macros, also for the same reason the bed mesh center_mesh cannot be called in the START_PRINT directly and need to be in the CALIBRATE_Z function, etc...

config/hardware/ercf.cfg
Frix-x: Oh well, symlinks... Ok this change a little bit the thing yes... I'll try to think about how to do it properly :)

macros/base/end_print.cfg
```diff
@@ -16,14 +16,12 @@ gcode:
...
             # unload filament and park into ercf
             ERCF_EJECT
-        {% elif unload|int == 0 %}
-        {% elif unload|int > 1 and ercf_unload_on_end_print %}
+        {% elif unload|int == 1 %}
             # unload filament and park into ercf
             ERCF_EJECT
-        {% else %}
         {% endif %}
     {% else %}
```
Frix-x, line 26: I think it's too complicated.... As stated in another comment: slicer should be used in prioritary as it is for the other parameters.

Juste use something like :
``` 
# Set the parameter like that
{% set ercf_unload = params.ERCF_UNLOAD_AT_END|defaultprinter["gcode_macro _USER_VARIABLES"].ercf_unload_on_end_print)|int %}

# Then just use it in an easy way like that
[....]
{% if ercf_enabled and ercf_unload   %}
    ERCF_EJECT
{% else %}
[....]
```

Surion79: basically it does what you wanted, just not as clean as your solution :)

Surion79:
> > What about a M400 and then check if printer.ercf.filament == "Loaded" and if not loaded then pause? Or even add a retry?
> 
> No it will not work. I think you miss the understanding of how template expansion is working (the Jinja). So just a little bit of explanation: when you run a macro there is a two step process happening in Klipper:
> 
> 1. Klipper read the Jinja to create the gcode internally. This means that all the variables used in the Jinja are read at this specific instant. If you have for example a "for" loop to move axis, then it will just create the corresponding Gcode and replace it directly. All the "loops" are created at this moment
> 2. Then when the template expansion is finished, there will be only standard gcode with all the values replaced, etc... This gcode will be executed with no "intelligence", just line by line. So no "specific checks" can be done at this moment.
> 
> At the time we wrote the ERCF error macro, the `START_PRINT` was just a monolitic big macro. So at START_PRINT call, during the template expansion phase the ERCF was ok and all the gocde generated properly. Then during the START_PRINT procedure, if for some reason the ERCF was erroring, as the template expansion was already done, there was nothing happening as the variable was not even read. The solution for that is to call another macro to do the check as this new macro will have its own context.
> 
> It's also for the same reason that `HEATSOAK_CHAMBER` and `_WAIT_CHAMBER_TEMP` is done in two macros, also for the same reason the bed mesh center_mesh cannot be called in the START_PRINT directly and need to be in the CALIBRATE_Z function, etc...

but that applies to the 2nd of my two questions?
i think it should be possible to check a variable against a constant?

Benoitone:
> but that applies to the 2nd of my two questions?
> 
> i think it should be possible to check a variable against a constant?

Unfortunately no, the variable keeps its value in memory which is defined at the beginning of the gcode and the state of it only changes when a new gcode is called. It must therefore be put in another gcode.
And as the purpose of the klippain start_print is to be modular, at what stage put it??

Surion79:
> > but that applies to the 2nd of my two questions?
> > i think it should be possible to check a variable against a constant?
> 
> Unfortunately no, the variable keeps its value in memory which is defined at the beginning of the gcode and the state of it only changes when a new gcode is called. It must therefore be put in another gcode. And as the purpose of the klippain start_print is to be modular, at what stage put it?? Maybe _MODULE_PURGE ?? but I don't know if everyone use purge bucket...

So we need a function call to update the value of a variable by using it in the function?  I guess this is the trick? (I don't mean as parameter, since that would only inherit the value from the calling function)
When I get an answer to this, I think I can implement an idea

Benoitone:
Yes, but even if the value of the variable is up to date, the only way I found to stop the macro that calls the test is to do action_raise_error that actually displays a triple message.
I don't quite understand why you don't want to leave it in place? And if you find a cleaner solution you will change it. But I had done a lot of tests in different ways even trying to recover the variables after calling the test gcode and it didn't work to stop gcode...

And in the event that you manage to pause your start print how do you reload (technically?) The filament that, for one reason or another, was not loaded (generally because it was not taken by the gears of the ecf or because of a bad servo position)?

Frix-x:
Yeah I think this function is ok. I don't understand what is the problem with that?

Also as a side note, it was mandatory in the past when Klippain v2 was using a monolitic `START_PRINT` macro. So if you started it, there was no way to check during the process and we needed to wait until the end. That's why it was done this way: to call another macro inbetween to do the check and stop.
Today in Klippain v3, there is the `START_PRINT` sequencer built-in. This sequencer call different sub-macros for each actions. So this means that we could theoretically check the variable at the begining of every actions and stop if there is an error (to account for the different order user could use). Be this is maybe not the best way and I would prefer to check after the filament load as it's done today to avoid duplicates.

Surion79:
> And in the event that you manage to pause your start print how do you reload (technically?) The filament that, for one reason or another, was not loaded (generally because it was not taken by the gears of the ecf or because of a bad servo position)?

as option it is nice, so you don't have to do all the print init (heating, QGL, bed mesh) again, because you can then just do ercf_unlock, load (different) filament and continue.

Benoitone:
Do you think it's possible to pause the current gcode, execute manually other macros and go back just after the pause?

Surion79:
> Do you think it's possible to pause the current gcode (start_print), execute manually other macros and go back just after the pause?

That is a good question I haven't thought about. Do you know the answer to this, @Frix-x?

Surion79:
To prevent any more delay on the v4 release, i uncommened the line and we can pick it up later if there is a need for

Surion79 (commented):
prep for undrafting

Surion79:
i think this can get a final review and be done in v4 develop imho

Benoitone:
To update docs/configuration.md, by default you prefer:
`variable_ercf_unload_on_end_print: True or False` ?

Surion79:
> To update docs/configuration.md, by default you prefer: `variable_ercf_unload_on_end_print: True` or `False` ? Actually I make a mistake...

true would be my recommendation

Surion79 (commented):
@Benoitone Outstanding work, i believe it took some time to find all these "spaces"

Benoitone:
> @Benoitone Outstanding work, i believe it took some time to find all these "spaces"

Yes, I had trouble sleeping... it kept me a little busy.

macros/base/start_print.cfg
```diff
@@ -325,9 +347,11 @@ gcode:
 
     # Load intial tool if needed
     {% if ercf_enabled %}
-        ERCF_RESET_STATS
-        ERCF_CHANGE_TOOL_STANDALONE TOOL={INITIAL_TOOL}
-        _DO_ERROR_ERCF_PAUSE
+        {% if ercf_reset_stats_on_start_print %}
+            ERCF_RESET_STATS
```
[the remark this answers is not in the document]
Frix-x: Also can this be moved directly in the top of the `START_PRINT` macro before calling any module? I feel like this should not stay here as if for example someone tweak its `START_PRINT` sequence and do not include an extruder heating phase, this will never happens even if it's set to true in the `variables.cfg`

macros/base/start_print.cfg
```diff
@@ -325,9 +347,11 @@ gcode:
...
+        {% endif %}
+        ERCF_CHANGE_TOOL STANDALONE=1 TOOL={INITIAL_TOOL}
+        _DO_ERROR_ERCF_PAUSE
```
Frix-x, line 354: Can this macro be renamed in: `_ERCF_ERROR_CHECK` ? It would help understand the problem and feature adressed by this custom macro

install.sh
```diff
@@ -222,6 +222,7 @@ function install_mcu_templates {
             filename=$(basename "${file_list[$((ercf_template-1))]}")
             cat "${FRIX_CONFIG_PATH}/user_templates/mcu_defaults/ercf/$filename" >> ${USER_CONFIG_PATH}/mcu.cfg
             printf "[CONFIG] Template '$filename' inserted into your mcu.cfg user file\n\n"
```
Frix-x, line 224: As you added a new line just after this one, you will also need to change it regarding the `\n` at the end of the line.
So can you change this line to:
```
echo "[CONFIG] Template '$filename' inserted into your mcu.cfg user file"
```

macros/miscs/debugging.cfg
```diff
@@ -35,3 +35,19 @@ gcode:
             {% endfor %}
         {% endif %} 
     {% endfor %}
+
+# Specific macro to raise an error during the print_start and stop in case of ERCF error
+[gcode_macro _DO_ERROR_ERCF_PAUSE]
```
Frix-x, line 40: Can this macro be renamed in: _ERCF_ERROR_CHECK or something like that? It would help understand the problem and feature adressed by this custom macro

macros/miscs/debugging.cfg
```diff
@@ -35,3 +35,19 @@ gcode:
...
+
+    {% if printer.ercf.enabled|int == 1 %}
+        {% if printer.ercf.tool|int != INITIAL_TOOL or printer.ercf.filament != "Loaded" %}
+            {action_raise_error("ERROR in loading filament from ERCF !!! Check and restart the print.")}
+        {% else %}
+            {% if verbose %}
+                RESPOND MSG="T{INITIAL_TOOL} ready"
+            {% endif %}
+        {% endif %}
+    {% endif %}
```
Frix-x, line 53: Can you add a free new line at the end of the file?
Also I feel that this macro definition should be moved directly into the `ercf.cfg` file as it's used only when there is an ERCF connected.

macros/miscs/startup.cfg
```diff
@@ -103,6 +108,14 @@ gcode:
     {action_respond_info(parameters.output)}
 
 
+[gcode_macro _INIT_ERCF]
+gcode:
+    {% set ercf_enabled = printer["gcode_macro _USER_VARIABLES"].ercf_enabled %}
+    {% if ercf_enabled and printer.ercf.enabled|int != 1 %}
```
Frix-x, line 114: As it's already checked if `ercf_enabled` is true in the `KLIPPAIN_STARTUP` macro, you can remove it from here and just use instead:
```
{% if printer.ercf.enabled|int != 1 %}
```

macros/miscs/startup.cfg
```diff
@@ -103,6 +108,14 @@ gcode:
...
+        RESPOND MSG="ERCF is set as enabled but supported ERCF software "Happy Hare" has not been identified. Please check your Happy Hare installation."
```
Frix-x, line 115: This comment need to be changed to an error and can be improved like:
```
{ action_raise_error("ERCF is enabled in the Klippain, but "Happy Hare", the supported ERCF backend software is not detected. Please install it now from https://github.com/moggieuk/ERCF-Software-V3 !") }
```

user_templates/mcu.cfg
```diff
@@ -12,4 +12,3 @@
 # [board_pins] section. You can use the pinout documentation here: ./docs/pinout.md
 
 # ------------------------------------------------------------------------------------------
-
```
Frix-x, line 15: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates inside from the install script

user_templates/mcu_defaults/ercf/Fysetc_ERCF_ERB.cfg
```diff
@@ -34,4 +34,3 @@ aliases:
     EXTRA_PINS9=MCU_ADC0      , EXTRA_PINS10=MCU_ADC1     ,
     EXTRA_PINS11=MCU_ADC2     , EXTRA_PINS12=MCU_ADC3     ,
     EXTRA_PINS13=<GND>        , EXTRA_PINS14=<5V>         ,
-
```
Frix-x, line 37: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/ercf/Mellow_fly_ERCF.cfg
```diff
@@ -35,4 +35,3 @@ aliases:
     EXTRA_PINS9=MCU_P28  , EXTRA_PINS10=MCU_P12 ,
     EXTRA_PINS11=MCU_P29 , EXTRA_PINS12=MCU_P24 ,
     EXTRA_PINS13=MCU_P25 , EXTRA_PINS14=MCU_P13 ,
-
```
Frix-x, line 38: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/ercf/Tircown_ERCF_easy_brd.cfg
```diff
@@ -25,4 +25,3 @@ aliases:
     ERCF_SERVO=MCU_SERVO ,
 
     ERCF_ENCODER=MCU_ENCODER ,
-
```
Frix-x, line 28: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/main/BTT_Octopus.cfg
```diff
@@ -44,4 +44,3 @@ aliases:
     LIGHT_OUTPUT=MCU_HE2         ,
     LIGHT_NEOPIXEL=MCU_STOP5     ,
     STATUS_NEOPIXEL=MCU_NEOPIXEL ,
-
```
Frix-x, line 47: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v2.cfg
```diff
@@ -52,4 +52,3 @@ uart_address: 1
 uart_pin: TMCUART
 tx_pin: TMCTX
 uart_address: 3
-
```
Frix-x, line 55: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/main/BTT_SKR_E3mini_v3.cfg
```diff
@@ -52,4 +52,3 @@ uart_address: 1
 uart_pin: TMCUART
 tx_pin: TMCTX
 uart_address: 3
-
```
Frix-x, line 55: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/main/BTT_SKR_Pico_v1.0.cfg
Frix-x, line 1: I think this file also need another free new line at the end to be injected like the other as well. Since there is also a lot of changes like that in this commit, can you add it here as well?

user_templates/mcu_defaults/main/Double_BTT_SKR_v1.4.cfg
Frix-x, line 1: I think this file also need another free new line at the end to be injected like the other as well. Since there is also a lot of changes like that in this commit, can you add it here as well?

user_templates/mcu_defaults/main/Fysetc_Spider_v1.x.cfg
```diff
@@ -44,4 +44,3 @@ aliases:
     LIGHT_OUTPUT=MCU_E1_OUT         ,
     LIGHT_NEOPIXEL=MCU_RGB_B        ,
     STATUS_NEOPIXEL=MCU_5V_RGB_DATA ,
-
```
Frix-x, line 47: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/mcu_defaults/main/Fysetc_Spider_v2.x.cfg
```diff
@@ -44,4 +44,3 @@ aliases:
     LIGHT_OUTPUT=MCU_E1_OUT         ,
     LIGHT_NEOPIXEL=MCU_RGB_B        ,
     STATUS_NEOPIXEL=MCU_5V_RGB_DATA ,
-
```
Frix-x, line 47: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/moonraker.conf
```diff
@@ -34,4 +34,3 @@
 
 ##### Add your custom moonraker config customizations and overrides below this line...
 # ------------------------------------------------------------------------------------
-
```
Frix-x, line 37: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/overrides.cfg
```diff
@@ -63,4 +63,3 @@
 
 # [stepper_z3]
 # dir_pin: !Z3_DIR
-
```
Frix-x, line 66: Can you put back this second free new line at the end? It was there on purpose to allow better separation when injecting the templates from the install script

user_templates/variables.cfg
```diff
@@ -171,6 +171,15 @@ variable_material_parameters: {
...
 
 
+################################################
+## ERCF 
+################################################
+## this section is only considered if an ercf is installed and configured
+
+variable_ercf_unload_on_cancel_print: False
+variable_ercf_unload_on_end_print: True
+variable_ercf_reset_stats_on_start_print: False
```
Frix-x, line 181: Check the discutions in the previsous comment by I'm quite unsure if we need all these variables

user_templates/printer.cfg
```diff
@@ -166,6 +166,12 @@
 
 # ---------------------------------------------------------- ERCF ----
 ### ------------------------------------------------------------------
+### Klippain is optimized to be used with ERCF Happy Hare https://github.com/moggieuk/ERCF-Software-V3
+### All this files created by install script of Happy Hare must be include
```
Frix-x, line 170: "All these files are created by the Happy Hare install script and must be included"

Frix-x (dismissed):
Hellooo !

Ok, so I took some time to do a full review on this PR in order to merge it today if possible in Klippain :) I think we are almost ready!
You'll see that there is lot of comments (like ~35) with a lof of changes asked. But don't panic, most of them are cosmetic changes to streamline the ERCF integration and make it better aligned with how the rest of Klippain is designed :)

Also as a side note, thank you very much @Benoitone for all the spaces and tabs cleanup, this is something I planned to do for a while. So thank you for that. However, next time, please do that in an another PR to avoid mixing everything in one, it would help to keep things clean and focus changes topic by topic in every PRs :)

config/hardware/ercf.cfg
```diff
@@ -1,158 +1,12 @@
...
 [gcode_macro _USER_VARIABLES]
 variable_ercf_enabled: True
 gcode:
 
-# Include the ERCF macros to unlock them when this file is
-# included in the printer.cfg config file
-[include ../../macros/miscs/ercf.cfg]
+[extruder]
+max_extrude_only_distance: 200
+max_extrude_cross_section: 50.0
```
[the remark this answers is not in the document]
Surion79: Primeline still works. 50 is required by ercf or it doesn't work 😅
I added it in the ercf.cfg. It is copied from the manual. So no manual work needs to be done.

macros/base/cancel_print.cfg
```diff
@@ -5,16 +5,17 @@ gcode:
     {% set turn_off_heaters_in_end_print = printer["gcode_macro _USER_VARIABLES"].turn_off_heaters_in_end_print %}
     {% set safe_extruder_temp = printer["gcode_macro _USER_VARIABLES"].safe_extruder_temp|float %}
     {% set ercf_enabled = printer["gcode_macro _USER_VARIABLES"].ercf_enabled %}
+    {% set ercf_unload_on_cancel_print = printer["gcode_macro _USER_VARIABLES"].ercf_unload_on_cancel_print %}
```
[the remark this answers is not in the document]
Surion79: I prefer have the option to have both. Cancel is an irregular print end and it should be possible to be able to control this circumstance. It doesn't hurt and the variables are clean to understand

config/hardware/ercf.cfg
```diff
@@ -1,158 +1,12 @@
...
-# configurable, but fairly fixed values
-gear_stepper_accel: 0                                                              # The acceleration value applied to the gear stepper on moves, standard is to use 0
-timeout_pause: 72000                                                               # Time out used by the ERCF_PAUSE
-disable_heater: 600                                                                # Delay after which the hotend heater is disabled in the ERCF_PAUSE state
-min_temp_extruder: 180                                                             # Temp used during the auto-calibration macro, to ensure we can move the extruder (but not really extruding)
-unload_buffer:50                                                                   # The amount, in mm, to reduce the unload so that the more accurate encoder unload has room to operate
-extruder_homing_max: 100                                                           # the maximum distance to advance in order to attempt to home the extruder
-extruder_homing_step: 2                                                            # the step size to use when homing the extruder
+[pause_resume]
+recover_velocity: 300.0
```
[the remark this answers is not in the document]
Surion79: removed section in ercf.cfg

user_templates/variables.cfg
Surion79: i do believe that we need them.
Why would Moggieuk and EtteGit write statistics in a persistent file when it is erased at begin of every print?..........

macros/base/start_print.cfg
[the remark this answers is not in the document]
Benoitone: ERCF_RESET_STATS is usefull for reset the statistics like numbers of swap in the current print... Personnaly i'll prefer to make a reset at the beginning of the print, so it cleans the previous statistics.

user_templates/printer.cfg
Surion79: done

macros/base/start_print.cfg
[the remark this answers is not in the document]
Surion79: thats personal choice. Ette und Moggieuk write statistics in a persistent file to be loaded on startup. I don't think they do that so the statistics are deleted on every print.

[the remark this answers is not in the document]
Surion79: > Also can this be moved directly in the top of the `START_PRINT` macro before calling any module? I feel like this should not stay here as if for example someone tweak its `START_PRINT` sequence and do not include an extruder heating phase, this will never happens even if it's set to true in the `variables.cfg`

You really do not own an ERCF.
ERCF loading means heating the extruder to at a configured value, which is at least 200 degrees which would make an issue with tap. Benoitone and i run this with ercf in exact this configuration successfully.

Surion79:
@Benoitone Don't modify the start print regarding thje clog detection. I think @Frix-x has not understood the meaning of the coding there.

macros/base/start_print.cfg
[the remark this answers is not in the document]
Surion79: I wouldn't change it.

Benoitone:
I have a question... Why using `{% if ercf_enabled %}` instead of `{% if printer.ercf.enabled %}`?
it would allow the user to use this feature of HH

ercf

macros/miscs/startup.cfg
Benoitone: can we replace `{% if printer.ercf.enabled|int != 1 %}` by `{% if printer.ercf.enabled %}`?
this variable introduce by HH is boolean (True or False)

macros/base/start_print.cfg
```diff
@@ -190,7 +201,17 @@ gcode:
     G1 E-0.2 F2100
     G92 E0
     G1 Z3 F{Sz}
+    M400
 
+    {% if ercf_enabled %}
+        {% if clog_detection > 0 %}
```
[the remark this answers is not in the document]
Benoitone: Clog detection is not reset when ERCF is disabled and as 4 different values

Surion79 (commented):
looks good

Surion79:
> I have a question... Why using `{% if ercf_enabled %}` instead of `{% if printer.ercf.enabled %}`? it would allow the user to use this feature of HH
> 
> ercf

Can we please put this on the stash for later and finish this PR and get the release v4 going? It is already really big.

macros/base/start_print.cfg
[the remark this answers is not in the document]
Frix-x: > You really do not own an ERCF. ERCF loading means heating the extruder to at a configured value, which is at least 200 degrees which would make an issue with tap. Benoitone and i run this with ercf in exact this start function successfully.

I was speaking about the resest stats only

Benoitone:
I add some {% if printer.ercf.enabled|int == 1 %} to prevent issue if user make ERCF_DISABLE.
I'm not sur if |int == 1 is necessary...

Frix-x:
Please don't touch this branch from now. I'll cleanup the code a little bit for the merge :)

Benoitone:
> > I add some {% if printer.ercf.enabled|int == 1 %} to prevent issue if user make ERCF_DISABLE. I'm not sur if |int == 1 is necessary...
> 
> 
> 
> I don't agree with your changes. you added another layer of ifs into this. Have you tested everything? Any changes on that level require testing the integration completely again. This PR is never gonna finish if you add something "small". At some point its end of development and testing only. I explicitly asked you regarding this and instead of replying to my message you just add code.

I have test it yesterday evening. It works with my print in progress. I didn't test if using ERCF_DISABLE previously...
I've change too the reactivate clog detection to be compatible with yours changes in desactivate

Frix-x:
So I just pushed a new commit with some change regarding this. Now it should work universaly as we first check for the klippain ERCF include before looking at the HH enabled state (it should be done in this order as the opposite can lead to bug if HH is not installed like on my machines)

I also added the same clog detection enable/disable logic in in the other place where it should be needed (loading/unloading filament manually and END_PRINT tip shaping).

Frix-x:
So I just pushed a new commit with some change regarding this. Now it should work universaly as we first check for the klippain ERCF include before looking at the HH enabled state (it should be done in this order as the opposite can lead to bug if HH is not installed like on my machines)

I also added the same clog detection enable/disable logic in in the other place where it should be needed (loading/unloading filament manually and END_PRINT tip shaping). I do know that it should not be needed in the UNLOAD_FILAMENT and LOAD_FILAMENT as the ERCF is having its own internals, but it's in case someone call them mannually. Maybe this should be checked and use ERCF_EJECT and ERCF_LOAD instead?

Frix-x:
If everything is ok for you now, I think we can merge it to the develop branch. Please do not do any modification from now on.

Frix-x (approved):

Frix-x:
> Last question.. In ercf_software.cfg from ERCF HH there is a [save_variables] file define...
> 
> Because the ERCF files are loaded before variables.cfg in printer.cfg. is this entry [save_variables] from klippain really override the one from HH?

I've not tested it (as I don't have any ERCF nor HH installed), but I do not see why it would not work as the other overrides. It should be ok!

Benoitone:
> > Last question.. In ercf_software.cfg from ERCF HH there is a [save_variables] file define...
> 
> > 
> 
> > Because the ERCF files are loaded before variables.cfg in printer.cfg. is this entry [save_variables] from klippain really override the one from HH?
> 
> 
> 
> I've not tested it (as I don't have any ERCF nor HH installed), but I do not see why it would not work as the other overrides. It should be ok!

Test OK 👍

Benoitone:
Should we specify somewhere to do not update printer.cfg at the end of the ERCF HH install script?
script

Frix-x:
I can add a note in the configuration documentation for this :)

PR #203, opened 2023-05-11
Bed Fans draft PR for discussion

Surion79:
@Frix-x Could you please look at the macro overrides if these work with klippain? It looks very good. It is provided by Ellis Voron mods github.
I get my fans tomorrow and i think i can test it this weekend. but please look at it. it is very technical. i hope this works with chambersoak


Surion79:
Closing this PR as it is not really necessary if filters are used and if no chamber thermistor is available

PR #204, opened 2023-05-14
Case light off when printing  at define layer + command to turn On/Off case light

Benoitone:
This PR adds an option allowing:
- add gcode:
to turn on `SET_CASELIGHT_LED_ON` and turn off: `SET_CASELIGHT_LED_OFF` the case light of the frame.
- And also the possibility during printing to turn off the caselight from a certain layer (X); by default set to layer 2
To be done you need to add the following commands in the before layer change gcode of the slicer (SuperSlicer exemple):
`SET_PRINT_STATS_INFO CURRENT_LAYER={layer_num}
_STOP_CASELIGHT_LAYER_CHANGE LAYER=X`
where X is the value of the layer.

### DIFF
--- a/config/hardware/lights/fcob_white.cfg
+++ b/config/hardware/lights/fcob_white.cfg
@@ -1,6 +1,7 @@
 # White lights using 24V (example: white fcob)
 [gcode_macro _USER_VARIABLES]
 variable_light_enabled: True
+variable_caselight_off_layer_enabled = True
 variable_light_pin_name: "caselight"
 gcode:
 
--- a/config/hardware/lights/fcob_white.cfg
+++ b/config/hardware/lights/fcob_white.cfg
@@ -7,6 +7,7 @@ gcode:
 
 # Also include directly the white lights control macros from here
 [include ../../../macros/hardware_functions/caselights.cfg]
+[include ../../../macros/miscs/framelight.cfg]
 
 
 [output_pin caselight]
--- a/config/hardware/lights/fcob_white.cfg
+++ b/config/hardware/lights/fcob_white.cfg
@@ -1,7 +1,6 @@
 # White lights using 24V (example: white fcob)
 [gcode_macro _USER_VARIABLES]
 variable_light_enabled: True
-variable_caselight_off_layer_enabled = True
 variable_light_pin_name: "caselight"
 gcode:
 
--- a/config/hardware/lights/neopixel_caselight.cfg
+++ b/config/hardware/lights/neopixel_caselight.cfg
@@ -1,6 +1,7 @@
 # Neopixel leds used as general printer lights
 [gcode_macro _USER_VARIABLES]
 variable_status_leds_caselight_enabled = True
+variable_caselight_off_layer_enabled = True
 variable_status_leds_caselight_led_name: "caselight"
 gcode:
 
--- a/config/hardware/lights/neopixel_caselight.cfg
+++ b/config/hardware/lights/neopixel_caselight.cfg
@@ -7,6 +7,7 @@ gcode:
 
 # Also include directly the leds control macros from here 
 [include ../../../macros/hardware_functions/status_leds.cfg]
+[include ../../../macros/miscs/framelight.cfg]
 
 
 [neopixel caselight]
--- a/config/hardware/lights/neopixel_caselight.cfg
+++ b/config/hardware/lights/neopixel_caselight.cfg
@@ -1,7 +1,6 @@
 # Neopixel leds used as general printer lights
 [gcode_macro _USER_VARIABLES]
 variable_status_leds_caselight_enabled = True
-variable_caselight_off_layer_enabled = True
 variable_status_leds_caselight_led_name: "caselight"
 gcode:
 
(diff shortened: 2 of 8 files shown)


Surion79:
what is the exact use case for this? Does this overwrite timelapse settings from slicer?

Benoitone:
> what is the exact use case for this? Does this overwrite timelapse settings from slicer?
> 
> Would really love to have some sort of description within a PR what the function does and what it is for (use case)

It's too late to make a correct description :(
I have a problem: I can't get the values ​​of the current layer without going through a `SET_PRINT_STATS_INFO CURRENT_LAYER={layer_num}` in the Superslicer.
`{% set cur_layers = printer.print_stats.info.current_layer|int %}` still returns none when printing.
**can someone help me to retrieve this value directly?**

Surion79:
> can someone help me to retrieve this value directly?

would love to understand first what you want to achieve with this PR. It is code and no description what it does. And reading source code is different than understanding what you want to achieve. Because it could be also be done differently what you haven't thought about :)

Surion79:
Do you want to turn of case light when a layer of X has been reached?
Why not make it generic to be configurable for all possible lights and not only on/off, but possibly also dim for a variable value?

Benoitone:
> Do you want to turn of case light when a layer of X has been reached?
> 
> Why not make it generic to be configurable for all possible lights and not only on/off, but possibly also dim for a variable value? 
> 
> 

Yes, exactly, I would like to stop case light after X layers with only STOP_CASELIGHT_LAYER_CHANGE LAYER=X on before or after layer change gcode in the slicer.
Mainsail displays the current layer/total layer when printing, so I thought there was a way to retrieve this value directly, without having to add SET_PRINT_STATS_INFO CURRENT_LAYER=<value> in the slicer

Surion79:
> > Do you want to turn of case light when a layer of X has been reached?
> > Why not make it generic to be configurable for all possible lights and not only on/off, but possibly also dim for a variable value?
> 
> Yes, exactly, I would like to stop case light after X layers with only `STOP_CASELIGHT_LAYER_CHANGE LAYER=X` on before or after layer change gcode in the slicer. Mainsail displays the current layer/total layer when printing, so I thought there was a way to retrieve this value directly, without having to add `SET_PRINT_STATS_INFO CURRENT_LAYER=<value>` in the slicer

wouldn't it be better to write a moduleset where you add only one gcode code to the slicer and rest is done in klippain?  so e.g. in slicer you say "onlayer_change" and then it checks for caselight and executes timelapse if it is activated.

Benoitone:
It's also possible but I think for Timelapse we need to install the plugin for moonraker and configure the slicer too for check layers?
So maybe no solution to find the current layer without adding slicer on layer_change command.
Isn't it possible to obtain the current layer otherwise? Like this?
image

Benoitone:
> Yes getting the current layer is not possible directly in Klipper. Mainsail do some extrapolation on the "Z height" and increments. But there is still the Z hop that alter it and make it very difficult to be "sure".
> 
> 
> 
> The current solution that you can find in the timelapse plugin is to use a specific "on layer change" macro in the slicer custom gcode. So unless we add it and complexify the Klippain user settings (new requirements in the slicer config), it will be quite hard to do. So I would say to avoid that for now until a better solution is found as it's only a "cosmetic" feature.

👍 maybe by removing `variable_caselight_off_layer_height = 2` in `variables.cfg` and set the default value directly in the gcode it could be an hidden future ?

Surion79:
@Benoitone if you begin to add such calls in Slicer software, I tend to think, that this is not really something for a generic approach like Klippain. 
Klippain does not forbid you to add your own files and functions :) just put an include in the overridees.cfg

Benoitone:
Effectively, you're right. It's not the goal, it's just that I think it's a good things to do not stay the frame light on for hours. 
And as indeed I found no other way than to add these 2 lines to the slicer it becomes less generic... I thought I had added the comment in the configuration.md. It was rather in the sense that there was no need to create an additional variable. And also rather transparent if you do not want to use the function.

Surion79:
> Effectively, you're right. It's not the goal, it's just that I think it's a good things to do not stay the frame light ON for hours. And as indeed I found no other way than to add these 2 lines to the slicer it becomes less generic... I said in the sense that there was no need to create an additional variable. And also rather transparent if you do not want to use the function. I find the possibility to turn ON/OFF the case light useful for users too

so it is like timelapse? you don't add it, nothing happens? I do wonder how this is compatible with timelapse. putting on the light for a picture on layerchange does create flashing, if you need light for timelapse

Benoitone:
No, it's not like Timelapse. Just turn of lights at a define layer if you need/want.
If you need light, like effectively for Timelapse just don't change anything on your before layer change gcode of the slicer... (I don't use Timelapse anymore)

Surion79:
> No, it's not like Timelapse. Just to turn off lights at a define layer if you need/want. If you need light, like effectively for Timelapse just don't change anything on your before layer change gcode of the slicer... (I don't use Timelapse anymore)

but it only works if you add your code to the slicer, right? Then it is exactly like timelapse?

Benoitone:
For Timelapse you need to additional install the moonraker plugin (not included in klippain). https://github.com/mainsail-crew/moonraker-timelapse

Surion79:
> Yes. But for Timelapse you need to additional install the moonraker plugin (not included in klippain). https://github.com/mainsail-crew/moonraker-timelapse

i just meant the integration of the command in slicer instead of "slicerless"

PR #206, opened 2023-05-14
Manta M8P v1.1

crazy-daddy:
Added config for Manta M8P v1.1 based on manual & wiring recommendations. Working for me, but I don't use all the pins, so please handle with care :-)


Frix-x:
Thanks for you contribution!
Since I'm currently working on the Manta V1.0 board pins, I'll merge your PR to a temporary branch and work on both templates myself to make them fit with the other files before releasing to main (hopefully alongside Klippain v4).

PR #213, opened 2023-05-26
Support for other kinematics (including Switchwire)

Frix-x:
This PR introduces support for other Kinematics in Klippain (such as CoreXZ for Switchwire or the standard Cartesian for a large range of printers).
@ndanyluk thank you for the preliminary work. Do not hesitate to react in this PR regarding the changes that I plan to do :)


W141-ID:
If you need any help regarding that, I have a switchwire running on klippain through overrides

Frix-x:
> If you need any help regarding that, I have a switchwire running on klippain through overrides

Sure. I just pushed a new commit that completes the feature.

It would be very good if you could try it and confirm that it works ok for you. There are a lot of changes, so I suggest you install the last templates and start from scratch to make sure everything is ok and there are no artifacts from your old config.

You'll find that now:
  - Axis are defined axis by axis independently to simplify it between different machine types
  - TMC configuration has been added and is done directly in the `mcu.cfg` template file.

@ndanyluk if you find time to try it out, let me know if it still works ok for you?

On my side I will try to do regression tests on my own machines and see if everything still works as it should :)
Please expect some bugs as I renamed and refactored a lot of files and stuff so I might have missed some things in the different includes... ^^

W141-ID:
I it possible to change to this branch through the install script or do I have to install it manually?

Frix-x:
> I it possible to change to this branch through the install script or do I have to install it manually?

Not with the installer directly. But you can do it this way:
  1. Just go into the klippain_config folder using SSH, then issue `git checkout multi-kinematics` to switch the repo to this branch
  1. Remove your `printer_data/config` directory (you can also just remove the `.VERSION` file to break the detection and force a reinstall of the template while getting an automated backup)
  1. Then, you can call the local install script by doing `~/klippain_config/install.sh`

config/hardware/axis/Y/TMC/TMC2209_V0specific.cfg
```diff
@@ -0,0 +1,19 @@
+# Y TMC2209 definition with some custom TMC parameters to reduce the
+# VFAs (Vertical Fine Artifacts) produced by the first V0 LDO motors kit
+
+
+# Including first the default XY TMC2209 file
```
Surion79, line 5: just Y here

macros/base/homing/homing_overide.cfg
```diff
@@ -63,7 +63,7 @@ gcode:
 
     {% endif %}
 
-    {% if probe_type_enabled == "dockable" %}
+    {% if probe_type_enabled == "dockable" or "dockable_virtual" %}
```
Surion79, line 66: review if statement if this really works. Looks uncommon.

macros/base/homing/homing_overide.cfg
```diff
@@ -216,26 +220,29 @@ gcode:
...
+        # if dockable probe as virtual endstop, then dock the probe
+        {% if probe_type_enabled == "vorontap" or "dockable_virtual" %}
             DEACTIVATE_PROBE
         {% endif %}
 
         G0 Z{homing_zhop} F{z_drop_speed}
     {% endif %}
 
-    {% if probe_type_enabled == "dockable" %}
+    {% if probe_type_enabled == "dockable" or "dockable_virtual" %}
```
Surion79, line 238: review if statement if this really works. Looks uncommon.

macros/base/probing/generic_probe.cfg
```diff
@@ -57,7 +57,7 @@ gcode:
     {% set Sz = printer["gcode_macro _USER_VARIABLES"].z_drop_speed * 60 %}
 
     # If a dockable probe is defined, then check and DETTACH the probe
-    {% if probe_type_enabled == "dockable" %}
+    {% if probe_type_enabled == "dockable" or "dockable_virtual" %}
```
Surion79, line 60: review if statement if this really works. Looks uncommon.

user_templates/mcu.cfg
```diff
@@ -13,3 +13,28 @@
...
+#   TMCs DRIVERS          #
+#-------------------------#
+
+# ---------------------------------------------------------------------- AXIS DRIVERS ----> optional, but only one line need to be selected
+### 1. X Drivers -------------------------------------------------------------------------
+# [include config/hardware/axis/X/TMC/TMC2209.cfg]
+# [include config/hardware/axis/X/TMC/TMC2209_V0specific.cfg]
+### 2. Y Drivers -------------------------------------------------------------------------
+# [include config/hardware/axis/Y/TMC/TMC2209.cfg]
+# [include config/hardware/axis/Y/TMC/TMC2209_V0specific.cfg]
```
Surion79, line 27: i would suggest to add a comment behind the V0 specific, since it is only meant for early LDO releases of the v0.1

Surion79 (changes requested):
I checked most parts. I recommend adding some checks in the startup so that only one line is used where one line is mentioned in the printer.cfg

user_templates/mcu.cfg
```diff
@@ -13,3 +13,28 @@
...
+### 3. Z Drivers -------------------------------------------------------------------------
+# [include config/hardware/axis/Z/TMC/TMC2209_1-Motor.cfg]
+# [include config/hardware/axis/Z/TMC/TMC2209_3-Motors.cfg]
+# [include config/hardware/axis/Z/TMC/TMC2209_4-Motors.cfg]
+# ----------------------------------------------------------------------------------------
+
+
+# ------------------------------------------------------------------- EXTRUDER DRIVER ----> optional, but only one line need to be selected
```
Surion79, line 35: i think a little bit of more explanation regarding toolhead vs mcu e tmc should be added

Frix-x: I don't think there is more explanation to add as when there is a toolhead MCU used and installed, it will inject some lines at the end that will automatically override the TMC pins. So this should be uncommented to select the extruder TMC and it doesn't matter if it's on the toolhead or the main MCU : it works for both the same way

Frix-x:
On my side I've done some regression tests and it looks to work as before (I just need to validate it with a full print).
I also do plan on trying the dockable_virtual configuration to see if it works correctly (it should be possible on my setup without too many changes)

What is the status of this on your side?

Surion79:
I would like to test it when it is on develop. Manual review was completed already

Frix-x:
I've added some fixes regarding the dockable_virtual which was quite dangerous in some cases (I got some crashes on my test machine). Now it should be safer.

I've also done some small refactoring to clean up the code by removing the dockable_virtual macro overrides and merging the changes into the "normal" file: it's now simpler and there's less code duplication.

It should be ready for a merge in develop now if you confirm it

Surion79 (approved):
looks good :)

PR #223, opened 2023-06-04
Fix Error in ERCF clog detection in primeline

Benoitone:
i think there is an error in check clog detection for reactivated it in prime line module??


Benoitone:
clog variables

Benoitone:
clog variables

Surion79:
i do have clog detection after prime line.
what is your point? you are refering to the config file, not the real loaded value.
This is works exactly as expected.

Benoitone:
Ho yes i Forget this... thanks

PR #239, opened 2023-06-19
fixed line start in /proc/cpuinfo

Surion79:
When i checked my startup, I found strange "unknown" system info printed for my genuine Pi4. When checking the file, I found the culprit:
image

The script checks for "Model name", instead of "Model" as shown in the screenshot. I removed the " name" section of "Model name". This should fix this issue and still be compatible for "Model name" since only "Model" is checked


Benoitone:
Does it show the correct model to you?
I have unknown system and unknown model too.
image

PR #244, opened 2023-06-23
Added comments on reading the graphs

garyd9:
Just some generic advice.


Frix-x (approved):
Thanks for this, as discussed on the Voron Discord, I'll merge it as you are perfectly right on this :)

PR #245, opened 2023-06-25
Update neopixel_caselight.cfg

smwoodward:
Was including the wrong macro, was using status macro and not caselight


mrstone14:
This update never made it into main.  patch-1 was deleted and the neopixel_caselight.cfg file still references the status macro instead of caselight macro

Issue #242, opened 2023-06-22
Modify bed mesh behavior to follow recent Klipper deprecations


Frix-x:
Adaptive bed mesh is currently using the relative_reference_index in order to get a proper offset but this parameter is now deprecated in  one of the last Klipper release.

The fix should be pretty easy since the needed computation are already done properly in the current macro.
This issue is just to keep track of that and not forget about it for the next Klippain release :)

PR #249, opened 2023-06-29
Bed mesh replace relative_reference_index with zero_reference_position

elpopo-eng:
First attempt to correct this issue (https://github.com/Frix-x/klippain/issues/242)

### DIFF
--- a/config/software/bed_mesh/bed_mesh_250mm.cfg
+++ b/config/software/bed_mesh/bed_mesh_250mm.cfg
@@ -11,12 +11,6 @@ gcode:
 
 
 [bed_mesh]
-# FOLLOW THE FORUMULA FOR relative_reference_index to be centered, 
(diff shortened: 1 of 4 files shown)


Frix-x:
Thanks for this. As disccussed, it will not work currently since the zero reference position parameter cannot be overriden when calling the macro.

I placed a comment in the Klipper PR to discuss it: https://github.com/Klipper3d/klipper/pull/6242

Surion79:
If i see correctly this would only work if a bed centered bed mesh is being used, instead of a calculated center where the center of the piece(s) is/are? 
I do wonder why Klipper deprecates overridable parameters and introduce new ones which cannot be overridden. hmmm

Frix-x:
I'll close this PR since the work was done directly in the develop branch in this commit: 79d5e241727f05bb4c541d9c5a61c69942c32470

There was more change needed in order to make it work. Today it's somewhat working but much simpler and we have to add the center probe at the end of the mesh as moving the zero_reference_position is still not possible... But at least adaptive mesh is working correctly :)

PR #254, opened 2023-07-12
Check on startup if variables are available

Surion79:
I added a new function in the startup to check variables. this is draft and sort of PoC. I would like to get some ideas, if it is possible to be more modular. Unfortunately I don't know if Jinja can do magic and use the value of a parameter as a variable name.
Then it would be easier to extend when no variables are introduced.

This does NOT check if the value of the variables are meaningful/valid. It checks the pure existance of the variables.

### DIFF
--- a/macros/miscs/startup.cfg
+++ b/macros/miscs/startup.cfg
@@ -15,6 +15,9 @@ gcode:
     # Dump the MCU version to the console for the Klippy log
     _INIT_MCU_VER
 
+    # Check if all required variables are set/available
+    _INIT_CHECK_VARS
+
     # Check if there is an extruder set
     {% set extruder_enabled = printer["gcode_macro _USER_VARIABLES"].extruder_enabled %}
     {% if not extruder_enabled %}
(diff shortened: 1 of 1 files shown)


Frix-x:
Hi, thanks for this draft!

It's indeed something I wanted to do. However I feel like it's not needed for most of the variables.
Here is how I has the idea to do it:
1. Generalize (if not already the case) the use of special variables in all the included files to be able to know when an option is enabled or not. This already pretty well managed.
1. Then in the macro, first of all, get these module enabled variables only. In the macro when can then do an "if module is enabled, then do things"
1. And only in the "when do things" part we should do the specific variable retrieval. This will avoid an error if the variable is not set by the user but the user don't care since he is not using the module. Also it will handle pretty well the breaking change since when we add a new module with new variables, if the module is not used by the user, the additional variables are also not needed.
1. Finally for the other "general used" variables that are not module dependant and used everywhere, we can add your solution to complete the handling :)

What do you think about that?

Surion79:
Checking the variables only means the one in the variables.cfg, or in general: all files not under the strict control of the repo. 

My approach was to check all variables in variables.cfg, but without check if they are really needed/used.
My idea behind it: start with a simple solution and go forward with it. 
My thoughts: 
- The "upgrade function" will come at some time, and then this work would be no longer needed. 
- Adding too much logic into this would even mean more complex maintenance just on this function. That is ok if only one person is working on the project, but this becomes complicated if multiple persons are working on a project. I can't add new stuff and then need to verify and check at which place i need to update the "check vars" function logic.

Open question still:  could we use jinja macros and use parameter values as variable names? THen the function itself would be a bit more complex, but the initial set up would be just a copied one liner. Or a function that can iterate through the default variables.cfg and matches with the current one of the user if variables of the current one are missing. that would be just a loop and even selfmaintaining.

PR #260, opened 2023-07-23
Default Filter time on end print

Surion79:
currently playing with my filters and found out, that there is no variable to configure the default duration for filter at the end print macro. this PR should address this "issue". the use of FILTER_TIME on END_PRINT is still possible and there is no change to that logic.

### DIFF
--- a/user_templates/variables.cfg
+++ b/user_templates/variables.cfg
@@ -182,6 +182,12 @@ variable_ercf_unload_on_cancel_print: False
 variable_ercf_unload_on_end_print: True
 variable_ercf_reset_stats_on_start_print: False
 
+################################################
+## Filter specific variables
+################################################
+## This section is only considered if a filter is available (and enabled)
+
+variable_filter_default_time_on_end_print = 600 # seconds
 
 ################################################
 # Other hardware options used in the macros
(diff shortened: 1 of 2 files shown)


user_templates/variables.cfg
```diff
@@ -182,6 +182,12 @@ variable_ercf_unload_on_cancel_print: False
 variable_ercf_unload_on_end_print: True
 variable_ercf_reset_stats_on_start_print: False
 
+################################################
+## Filter specific variables
+################################################
+## This section is only considered if a filter is available (and enabled)
+
+variable_filter_default_time_on_end_print = 600 # seconds
```
Frix-x, line 190: The variable name is not correct and will lead to an error

Frix-x (changes requested):

user_templates/variables.cfg
Surion79: What exactly is the error? Too long?

Frix-x: `variable_filter_default_time_on_end_print` is different than the one used in the macro : `variable_default_time_on_end_print`

Frix-x (approved):

PR #261, opened 2023-07-24
v4.1.0

Frix-x:


Frix-x:
I think we are ready for the v4.1.0

@Benoitone @Surion79 Do you want to wait for HHv2 and do a big release all at the same time or are you ok to split in two and focus v4.2.0 on the HHv2 topic ?

Surion79:
I would say to go 4.1 now and do HHv2 in 4.2
I think there is still some discussion open regarding compatibility to HHv1 and HHv2 in Klippain. I am not sure if klippain should force users to move from working HHv1 to HHv2. Especially since HHv2 is not officially released (repo says closed beta)

Frix-x:
It's indeed also my idea yeah!
So... I guess it's release day?

Frix-x:
Can you confirm everything is ok for you in the last state of the develop branch?
My machine is running fine currently also

Benoitone:
I think we can wait to merge HHv2... my branch is only compatible with V2 and V2 is also at beta state. Everything else is ok on my side. I think at this stage you can merge develop into main, and let a distinct branch for each one want to try HHv2 with Klippain.

Frix-x:
Let's go!

PR #266, opened 2023-07-30
Klipper restart handling on install

Surion79:
This PR is based on other repos how they handle post install procedures.


Surion79:
@Frix-x Did some work on it. CHecks now if interactive or not. if interactive, it asks if restart, if non-interactive, it restarts klipper

PR #269, opened 2023-07-31
added check for exclude_object availability

Surion79:
If Bed Mesh is called without a setting up exclude_objects, an error comes up. This is related to missing checks if something is (not) defined.
I added a variable and an additional check to fix the reported issue. The variable is required to prevent double else statements, which looks ugly


Frix-x (approved):
I added a change to how the check is done for the boolean since Jinja is pretty finicky with this due to how the Python True/False is done with the uppercase first letter. This makes "false" appears as "True" and is pretty weird. Using just a simple "if var" or "if not var" is more easy to understand and will avoid problems.

PR #270, opened 2023-08-03
Add switch sensor config

mattpolito:
As motion sensors will fail for a generic switch sensor due to not detecting the minimum amount of motion, I've provided a config and printer.cfg option for those who have a regular runout switch sensor.

ref: https://discord.com/channels/460117602945990666/1096701708643614820/1136286792920858794


Frix-x:
Hey, thanks for this PR !

I've got a couple of modifications to do to generalize the "enabled" variable. So I'll merge it in a temporary branch on my side and it will be released alongside Klippain next version :)

PR #273, opened 2023-08-06
Add support for sensorless homing on CoreXZ & Cartesian

ndanyluk:
Hi Frix!

New PR here to tweak the homing override such that CoreXZ behaves a little better with sensorless homing.  I've tested this PR on my VSW and it appears to be working nicely.

Specifically, I added a new variable to track kinematic type, then I use that variable in homing_override.cfg to control which motors have their current reduced during homing (no change for corexy).  Additionally, I increased the 1s wait to 2s between homing moves to allow stallguard registers to clear.  CoreXZ also required a 2s wait after the initial Z hop for the same reason.

Let me know if you need any changes or additional testing and I'd be happy to do it!

### DIFF
--- a/macros/base/homing/homing_override.cfg
+++ b/macros/base/homing/homing_override.cfg
@@ -3,6 +3,7 @@
 axes: xyz
 gcode:
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
+    {% set kinematics = printer["configfile"].config["printer"]["kinematics"] %}
     {% set probe_type_enabled = printer["gcode_macro _USER_VARIABLES"].probe_type_enabled %}
     {% set homing_zhop = printer["gcode_macro _USER_VARIABLES"].homing_zhop|float %}
     {% set homing_travel_speed = printer["gcode_macro _USER_VARIABLES"].homing_travel_speed * 60 %}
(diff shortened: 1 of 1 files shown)


Frix-x:
Hey thanks for this PR!

I'll take some time to review it today at the end of the day :)

Frix-x (changes requested):
Ok I've took some time to have a look. It's pretty good already but to make it a little bit more clean, I think you can use directly the Klipper configured kinematics instead of adding a new variable. What do you think about it?

Something like this should do the trick:
```
{% set kinematics = printer["configfile"].config["printer"]["kinematics"] %}
```

ndanyluk:
> Ok I've took some time to have a look. It's pretty good already but to make it a little bit more clean, I think you can use directly the Klipper configured kinematics instead of adding a new variable. What do you think about it?
> 
> Something like this should do the trick:
> 
> ```
> {% set kinematics = printer["configfile"].config["printer"]["kinematics"] %}
> ```

this is great I didn't realize I could do this! I'll implement and update the PR :)

ndanyluk:
@Frix-x ok that seems to work! Tested on my VSW with the suggested change and it still homes successfully/appears to set the currents properly for XZ, then Y

ndanyluk:
Ok I think I fixed it and squashed the commit history properly

Frix-x:
Thank you very much for this, it's a pretty cool feature :)

Frix-x (approved):

PR #275, opened 2023-08-08
Adding bed mesh for MK52 beds

ndanyluk:
Hi @Frix-x!

This PR is a minor update to enable bed mesh's on VSW and other Mk52-based printbeds :)

* adds a bed mesh config for the VSW's MK52 bed size (without zero_reference_index since VSW doesn't usually have one)
* slightly increases bed mesh's z hop between probes to account for larger probes (like the unklicky BFP which has a 12.6mm Z offset on my printer) on all bed_mesh configs. The increase was from 12->20mm.
* Amended the printer.cfg to offer this bed size as an option
* Updated overrides.cfg to add commonly changed probe settings (for non-TAP probes)

Let me know if you need me to change anything!
-Nick

### DIFF
--- a/config/software/bed_mesh/bed_mesh_120mm.cfg
+++ b/config/software/bed_mesh/bed_mesh_120mm.cfg
@@ -12,7 +12,7 @@ gcode:
 
 [bed_mesh]
 speed: 350
-horizontal_move_z: 12
+horizontal_move_z: 20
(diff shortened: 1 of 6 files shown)


Frix-x (approved):

Frix-x:
Thanks for the PR! I'll merge it :)

PR #279, opened 2023-08-09
Safer clean_nozzle macro

Fragmon:
The orbiter2 dont use a gear ratio.

https://trianglelab.net/u_file/2112/11/file/Orbiterv20FirmwareConfiguration-031c.pdf


Fragmon:
Protection from scratching the nozzle while traveling if the brush at the same level of the bed

Frix-x:
Thanks for the PR. I updated the target branch to develop since there was already some work done regarding the Orbiter configuration in the develop branch and it was conflicting. This PR now only focus on the Z hop added after purge/brush

Frix-x (approved):

PR #289, opened 2023-09-03
feat: add TMC5160 driver to X/Y

pvpacheco:
Adds TMC5160 driver option to X and Y axis configuration option. Those are used to run steppers at 48v.


Frix-x:
Hi thanks for this PR! I planned to add support for TMC5160 in the next release, so I'm pretty happy with this :)

Just to understand and find the best way to integrate it, can you tell me which MCU you are using?
Indeed, the SPI bus used for all the currently supported MCUs is the software SPI. Here is an example for the TMC2240 config files:
```
[tmc2240 stepper_x]
cs_pin: X_TMCUART
spi_software_sclk_pin: DRIVER_SPI_SCK
spi_software_mosi_pin: DRIVER_SPI_MOSI
spi_software_miso_pin: DRIVER_SPI_MISO
```

pvpacheco:
Hi, thank you for sharing klippain!

I'm using a Octopus Max EZ. Just changed to use software SPI and UART param. Its already working on my setup. 

What do you think?

Frix-x:
Thanks for this. I think it's ok for me, let's merge it :)

Frix-x (approved):

PR #290, opened 2023-09-09
remove usage of `homing_positive_dir`

Gethe:
As stated in [Klipper docs](https://www.klipper3d.org/Config_Reference.html?h=homing_positive_dir#stepper): "It is better to use the default than to specify this parameter. The default is true if `position_endstop` is near `position_max` and false if near `position_min`."

Including this parameter seems to be at best redundant and at worst potentially damaging for printers that home at axis minimum.

### DIFF
--- a/macros/base/homing/homing_override.cfg
+++ b/macros/base/homing/homing_override.cfg
@@ -315,10 +315,12 @@ gcode:
         G28 Z0
 
         G91
-        {% if printer["configfile"].settings["stepper_z"]["homing_positive_dir"] == False %}
-            G0 Z{homing_zhop} F{z_drop_speed} # small Z hop to avoid grinding the bed (as we should be close to Z0 right now)
-        {% elif printer["configfile"].settings["stepper_z"]["homing_positive_dir"] == True %}
-            G0 Z-{homing_zhop} F{z_drop_speed} # small Z move in the opposite direction to avoid staying on the endstop (not dangerous since we should be at Z max)
+        {% if (printer.toolhead.position.z < homing_zhop) %}
+            # small Z hop to avoid grinding the bed (as we should be at Z min right now)
+            G0 Z{homing_zhop} F{z_drop_speed}
+        {% else %}
+            # small Z move in the opposite direction to avoid staying on the endstop (not dangerous since we should be at Z max)
+            G0 Z-{homing_zhop} F{z_drop_speed}
         {% endif %}
         G90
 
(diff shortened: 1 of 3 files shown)


macros/base/homing/homing_override.cfg
```diff
@@ -315,10 +315,12 @@ gcode:
...
-        {% if printer["configfile"].settings["stepper_z"]["homing_positive_dir"] == False %}
-            G0 Z{homing_zhop} F{z_drop_speed} # small Z hop to avoid grinding the bed (as we should be close to Z0 right now)
-        {% elif printer["configfile"].settings["stepper_z"]["homing_positive_dir"] == True %}
-            G0 Z-{homing_zhop} F{z_drop_speed} # small Z move in the opposite direction to avoid staying on the endstop (not dangerous since we should be at Z max)
+        {% if (printer.toolhead.position.z < homing_zhop) %}
+            # small Z hop to avoid grinding the bed (as we should be at Z min right now)
+            G0 Z{homing_zhop} F{z_drop_speed}
+        {% else %}
+            # small Z move in the opposite direction to avoid staying on the endstop (not dangerous since we should be at Z max)
+            G0 Z-{homing_zhop} F{z_drop_speed}
```
Frix-x, line 323: This will not work in case some people use a homing_zhop of 0 and can lead to dangerous behavior. Using the `homing_positive_dir` is indeed necessary here.

Frix-x (changes requested):
Thanks for this PR.

I agree that hardcoding the homing dir is not necessary anymore today. If I remember correctly, it was a mandatory parameter a couple of year ago but now that there is an automatic direction extrapolation from the endstop position, I agree that it could lead to dangerous behaviors. Let's remove this :)

However, as stated in the next comment, using the computed value is still necessary for the Zhop after homing the Z axis. I'm using the "settings" value and not the "config" value. So this should work even if there is nothing specified in the config.

macros/base/homing/homing_override.cfg
Gethe: I can't think of any reason why anyone would do that, but I could wrap it in a check that z-hop is more than 0.

Gethe: On second thought, I just reverted this change to keep overall complexity to a minimum.

Frix-x:
Yeah the check is not necessary as this parameter is populated automatically by Klipper and it's this value that we retrieve (not the one from the config).

Thanks, I'll merge for the next version :)

PR #291, opened 2023-09-10
Update BTT_Octopus_Max.cfg

NanoMetere:
image
When using octopus max mcu definitions, klipper error appears and found a single comma missing. 
after adding comma, it works just fine
스크린샷 2023-09-10 114652


Frix-x:
Oh, you're right! This is probably a copy paste error. Thanks for the correction :)

Frix-x (approved):

PR #294, opened 2023-09-12
Remove invalid option extruder

LongDono:
extruder is a valid option for filament_motion_sensor, but not for filament_switch_sensor

	Option 'extruder' is not valid in section 'filament_switch_sensor runout_sensor'
	
	Once the underlying issue is corrected, use the "RESTART"
	command to reload the config and restart the host software.
	Printer is halted


Frix-x:
Oh, good catch. I'm not using one myself so didn't see this error. I'll merge to the develop branch for the next release :)

Frix-x (approved):

LongDono:
Happy to help but I'm new to this... Hence my deleting the branch while on mobile and then restoring it thinking that was a mistake. On the desktop, I can see it says it was safe to delete now. Thank you for your all your work on Klippain.

PR #302, opened 2023-09-18
Fix bug in hotend_heater_ctrl.cfg that allowed probing while nozzle is hot

### LINKED ISSUE #301: Error in hotted_heater_ctrl.cfg allows probing while nozzle is too hot with "fix_heaters_temperature_settle" set to true
### Klippain branch

- [X] I confirm using the main branch

### Version

4.1.1 Release

### Describe the bug and expected behavior

Same basic issue as #300, but in a different place. If you want me to start putting instances of this into a single issue, just close it out with a comment to that effect and I'll add this one and all future instances to #300 and I'll update the title to be more appropriate.

I thankfully didn't discover this the hard way, but noticed this while reviewing some of the more important safety macros. I put it to the test and sure enough, if you have a macro waiting for the ACTIVATE_PROBE macro to complete cooling down the nozzle, it doesn't wait. The M109 command in ACTIVATE_PROBE does not stop subsequent macro from being called and the tilting procedure continues without regard for the actual nozzle temperature.

This can lead to a situation where the nozzle is too hot and probes the bed anyway thus burning the PEI sheet or shattering glass beds. This can occur easily if a user sets safe_extruder_temp in _USER_VARIABLES to a value higher than 150C and running the preheat nozzle module in START_PRINT. This behavior would take all but the most experienced Klipper users off guard as they would expect to be protected be the ACTIVATE_PROBE macro.

This is a safety issue and I'll be putting up an PR shortly to at least hot fix the issue until something more elegant can be crafted.

### Additional information and klippy.log

Here's the snippet in question. The M109 command does not stop more macros from being called while temp is stabilizing.

 ```
  #If a Voron TAP probe is defined, then check the temperature and lower it if needed
    {% elif probe_type_enabled == "vorontap" %}
        SAVE_GCODE_STATE NAME=BEFORE_TAP_ACTION

        {% set ACTUAL_TEMP = printer.extruder.temperature %}
        {% set TARGET_TEMP = printer.extruder.target %}

        SET_GCODE_VARIABLE MACRO=ACTIVATE_PROBE VARIABLE=temperature VALUE={TARGET_TEMP}

        {% if TARGET_TEMP > tap_max_probing_temp %}
            { action_respond_info('Extruder temperature target of %.1fC is too high for TAP probing, lowering to %.1fC' % (TARGET_TEMP, tap_max_probing_temp)) }
            M106 S255 ; 100% the part cooling fan to help the extruder cooling
            M109 S{tap_max_probing_temp}
            M106 S0   ; Stop the part cooling fan
        {% else %}
            # Temperature target is already low enough, but nozzle may still be too hot
            {% if ACTUAL_TEMP > tap_max_probing_temp + 3 %}
                M106 S255 ; 100% the part cooling fan to help the extruder cooling
                TEMPERATURE_WAIT SENSOR=extruder MAXIMUM={tap_max_probing_temp}
                M106 S0   ; Stop the part cooling fan
            {% endif %}
        {% endif %}
    {% endif %}
```

austinrdennis:
Fixes #301.

This is inelegant, but this prevents the nozzle from touching the bed surface when too hot in all cases. Let me know if you think of something better or want to tweak it, but at least it can be fixed for now. I'm guessing at the dwell time of 5 min in a worst case scenario. It could probably be less, but I only have a Revo and that cools down quick so I assume other hotness have considerably more thermal mass than mine.

### DIFF
--- a/macros/base/probing/generic_probe.cfg
+++ b/macros/base/probing/generic_probe.cfg
@@ -35,6 +35,7 @@ gcode:
             { action_respond_info('Extruder temperature target of %.1fC is too high for TAP probing, lowering to %.1fC' % (TARGET_TEMP, tap_max_probing_temp)) }
             M106 S255 ; 100% the part cooling fan to help the extruder cooling
             M109 S{tap_max_probing_temp}
+            G4 P270000 ; Wait for nozzle to cool down before probing (5 min for reasonable worst case scenario of 300C)
             M106 S0   ; Stop the part cooling fan
         {% else %}
             # Temperature target is already low enough, but nozzle may still be too hot
--- a/macros/base/probing/generic_probe.cfg
+++ b/macros/base/probing/generic_probe.cfg
@@ -35,7 +35,6 @@ gcode:
             { action_respond_info('Extruder temperature target of %.1fC is too high for TAP probing, lowering to %.1fC' % (TARGET_TEMP, tap_max_probing_temp)) }
             M106 S255 ; 100% the part cooling fan to help the extruder cooling
             M109 S{tap_max_probing_temp}
-            G4 P270000 ; Wait for nozzle to cool down before probing (5 min for reasonable worst case scenario of 300C)
             M106 S0   ; Stop the part cooling fan
         {% else %}
             # Temperature target is already low enough, but nozzle may still be too hot
--- a/macros/helpers/bed_heater_ctrl.cfg
+++ b/macros/helpers/bed_heater_ctrl.cfg
@@ -6,12 +6,19 @@
 rename_existing: M190.1
 gcode:
     {% set S = params.S|float %}
+    {% set actual_temp = printer.heater_bed.temperature|float %}
 
     {% set fix_heaters_temperature_settle = printer["gcode_macro _USER_VARIABLES"].fix_heaters_temperature_settle %}
 
     {% if fix_heaters_temperature_settle %}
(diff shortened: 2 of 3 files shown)


austinrdennis:
Fixes the issue #301 now that I fully understand why that was happening. My testing shows the issue to be corrected, ready for merge and issue closeout.

Frix-x:
You are right, but please look at my answer in the original issue :)

austinrdennis:
The latest commit is my best attempt at implementing a solution that both fixes the bug and provides the functionality for low thermal inertia heaters. It works as intended in my testing, but I'm just one person with one machine so there may be edge cases I haven't thought of.

It's possible to stall the printer indefinitely (put it in a "busy" status) as this is currently implemented: If a M109/M190 command is issued with a S value that is less than the ambient air temperature in the room the printer is in, the printer will simply freeze up waiting for temperature to stabilize below room temp. I think we can agree that scenario is unlikely to happen in normal use and is pretty much harmless if it does occur. I think worst case is some cooked filament is a nozzle left for a long period in this state, but nothing more.

Let me know what you think and how I could improve on it! Thanks for taking time out of your day to go back and forth with me.

Frix-x:
Ok, I think this is now good to go for me :)
Thank you for you work and help in debugging this feature! I'll merge it for the next version of Klippain!

Frix-x (approved):

PR #305, opened 2023-09-22
Rampon firmware for Kusba and etc.

W141-ID:
I've added the config for Rampon firmware for the KUSBA. It uses the recommended setings from the KUSBA github. I've been using it for weeks now and it works fine on my end.


Frix-x (approved):
This is ok for me. Thanks for this PR :)

PR #307, opened 2023-09-25
Corrected typeof in variable `variable_filter_default_time_on_end_print` configuration

JerrettDavis:
`variables.cfg` incorrectly uses an equal sign (`=`) instead of a colon (`:`) for `variable_filter_default_time_on_end_print`. This has been corrected.


Frix-x (approved):
Wow. Thanks for this, I just don't understand how this was not seen before ^^
I'll merge for the next version :)

Issue #214, opened 2023-05-26
Please implement conditional QGL


rugolini:
Loving Klippain. But I'm still playing with the start sequence to have it right to my needs.
I usually print repeated sets and have no need to QGL every time.
Can you please add a variable "conditional_QGL" that allows the macro check if QGL is done and bypass it?

Surion79:
Isn't this already available with `force_homing_in_start_print` set to false?

Surion79:
Isn't this already available with `force_homing_in_start_print` set to false?

rugolini:
Hum.... good point. Maybe its not clear, at least for me, that this option is also for QGL... could not find it in the documentation or in the variables file as well.

rugolini:
I just confirmed... it execute the QGL with the force_homing_in_start=false.

Surion79:
Can you identify at which point?
Have you set the variable to not turn off motors at print end?

rugolini:
Yes... its variable_disable_motors_in_end_print: False

Surion79:
What about my first question? :)

rugolini:
"Can you identify at which point?" ... here it goes the console output for the 2nd print:
10:19 AM
File selected
10:19 AM echo: Material 'ABS' is used
10:19 AM pressure_advance: 0.040000 pressure_advance_smooth_time: 0.040000
10:19 AM echo: Heating up bed...
10:23 AM echo: Heatsoak bed, 1mn left...
10:23 AM echo: Bed temperature OK
10:37 AM echo: Chamber temperature OK !
10:37 AM echo: Nozzle cleaning...
10:38 AM Quad gantry leveling   <=========

rugolini:
Thanks Frix. Yeah, I think its important to have choice, after all each Voron is unique (examples: Kynematic vs standard mount, bed plate quality, Y backers, PLA vs ABS printing temperatures, etc.). 
Sorry, but I'm confused with your message: "as I already have the conditional homing .. I see no reason to add a variable to do a Force QGL"  ... I think you meant "I see no reason to NOT add a variable, right"?

Frix-x:
> Sorry, but I'm confused with your message: "as I already have the conditional homing .. I see no reason to add a variable to do a Force QGL" ... I think you meant "I see no reason to NOT add a variable, right"?

Yeah sorry, I wrote it too fast without reading me back. I edited my message.
Simple answer: let's add a conditionnal QGL as a possibility, but I'll not push it as the default behavior.

[...]

PR #309, opened 2023-09-25
v4.2.0

Frix-x:
# Klippain v4.2.0

## Special Highlight: Shake&Tune Module :rocket:
We're excited to introduce the Shake&Tune module in Klippain, a significant evolution from the original IS workflow scripts. This module represents a major step forward in our commitment to enhancing the Klippain experience and printer tuning with:

- :art: Stunning visual overhaul tailored for comprehensive analysis, making it easier than ever to understand your printer's dynamics and frequency profiles. There is now also an automatic peak detection algorithm, custom advices based on the graphs, printer's mechanical health indicator, damping ratio estimation, etc...
- :wrench: Hassle-free setup as everything is integrated in Klippain for all of you if you've set an accelerometer. Just run the macro and look at the results as you were doing with the old IS workflow. The good news is that this module is also now available outside Klippain to use it in another Klipper configuration easily [here](https://github.com/Frix-x/klippain-shaketune).
- :bar_chart: And an enhanced input shaping documentation is also available directly in the [K-Shake&Tune module documentation](https://github.com/Frix-x/klippain-shaketune/tree/main/docs). We've added numerous elements to guide you in interpreting the graphs, making it easier to optimize your printer’s performance.

### Configuration and Templates:
- Fixed a typo in SKR Pico template on motor pin definition
- Added a prefilled template for a custom MCU configurations, facilitating the initial installation of Klippain with an MCU not yet officially supported
- Added a template for the Fysetc Spider 3
- Added a template for the BTT EBB SB2209 CAN v1.0 with the RP2040 chip
- Rampon ADXLs support was added for easy input shaping setting since this is now a pretty common type of ADXL probe
- Support for the ADXL345 SHT V2.x integrated sensor added
- Support for the portable input shaper from Fysetc (RP2040 based ADXL) added
- Added support for servo in SKR Mini E3 MCU to align this template with the other templates
- Added configuration for BTT Manta M8P V2, expanding board support with newer products
- Added support for [Galileo2 Z blocks and Galileo2 Extruder](https://github.com/JaredC01/Galileo2) that are both very good products!
- Added a template for the Creality printers bed due to a different sensor than the one used on Keenovo heaters. This is done by a small refactoring of the bed templates files but the old one is still valid, so it's not a breaking change for now but this file will likely be deprecated in the future
- Added config files for 180mm printers such as the Micron+ for example

### Filament Sensors:
- Removed invalid option `extruder` in filament switch sensor config file ([#294](https://github.com/Frix-x/klippain/pull/294))
- Flush Klipper buffer before resetting filament sensor to ensure accurate readings and no remaining filament movements before reactivating the sensor after a purge or something like that

### General macros and scripts changes:
- Reworked input shaper scripts for enhanced performance and usability ([K-Shake&Tune](https://github.com/Frix-x/klippain-shaketune) module)
- Added support for Extruder motion vibrations measurements ([K-Shake&Tune](https://github.com/Frix-x/klippain-shaketune) module)
- Reset filter macro at the beginning of the `START_PRINT` sequence to avoid a delayed filter stop caused by a previsously finished print
- Refactored `plot_graph.sh` into a Python script for better performance and flexibility. This also allow to solve the path problem when the user was not "Pi". **If you've made an override about this, be sure to remove it!**
- Solved an old issue with conditional checks for QGL/Z_TILT ([#214](https://github.com/Frix-x/klippain/issues/214)). Now the tilting process should really be conditionnal as it was originally designed when setting `variable_force_homing_in_start_print` to false!
- Corrected a small typo in variable configuration for `variable_filter_default_time_on_end_print`
- Enhanced M109 and M190 logic with improved temperature settling ([#302](https://github.com/Frix-x/klippain/pull/302)) in order to avoid a bug when going from an high temperature to a lower one when the `variable_fix_heaters_temperature_settle` was set to true
- Improved the bed heatsoak sequence to use the current bed temperature (instead of bed temperature target) for a more accurate real time reading

### System and Installation:
- Added an auto uninstall script, simplifying the removal process for users that want to go back to their old configuration or that installed Klippain by mistake. This also allow at the same time to try Klippain very easily without taking too much risks
- Made tweaks to the installation script and updated `.gitignore` for better setup experience.
- Fixed issue with the path set in the gcode_shell_command for `system_info.py` script resolving startup errors when the user was not "Pi". **If you've made an override about this, be sure to remove it!**

### LED and Status Indicators:
- Fixed status LEDs for separate nozzle and logo LEDs, ensuring correct operation when having multiple Neopixel chains
- Prevented occasional LED flashing that could occurs before homing in the `START_PRINT` sequence

### Miscellaneous:
- Moved (and improved) the input shaper analysis documentation directly in the [K-Shake&Tune module repository documentation](https://github.com/Frix-x/klippain-shaketune/tree/main/docs). The old markdown docs now point to the new one to avoid problems and keep an easy access to them
- Enhanced `START_PRINT` sequence to allow a bed mesh profile loading from slicers using the `MESH=xxxxx` parameter in the custom START_PRINT Gcode. When using this, the adaptive bed mesh will be canceled and the specified mesh profile loaded instead. This can be usefull for calibration prints or if you just want to get the same static mesh for every prints.
- Passed all the `START_PRINT` raw parameters in the `_MODULE_CUSTOM1/3` in case you want to do something fancy in them based on your slicer parameters. Feel free to use them as you like :)


Frix-x:
I think we are pretty ready to merge this one now. Is it ok for everyone?

Surion79 (approved):
Naming the heater bed by manufacturer might be a bit cryptic for some printer users. There is no longer a "default".

Frix-x:
> Naming the heater bed by manufacturer might be a bit cryptic for some printer users. There is no longer a "default".

This is the beginning: the old file is not removed yet so it's still working. But as we have done it for the probes, the extruders or the ADXL boards for example, I feel that doing it for the other file is also a good thing to personalize more and let more choice for the users. It will also be a good base for a "configurator" that I plan to work on later on to help install Klippain :)

PR #312, opened 2023-09-29
Add BTT EBB SB2209 CAN v1.0 RP2040

Benoitone:


dpeart:
The pin settings for the RP2040 look correct to me.

dpeart (approved):
Looks good to me.

Frix-x (approved):

user_templates/mcu_defaults/toolhead/BTT_EBB SB2209_CAN_RP2040_v1.0.cfg
```diff
@@ -0,0 +1,77 @@
...
+[mcu toolhead]
+##--------------------------------------------------------------------
+canbus_uuid: change-me-to-the-correct-canbus-id
+##--------------------------------------------------------------------
+
+# If you want to override the wiring of the BTT EBB SB2209 CAN v1.0 RP2040, keep in mind that this
+# board is defined using the "toolhead" name. So you should use "pin: toolhead:PIN_NAME"
+# in your own overrides.cfg files.
+
+[include config/mcu_definitions/toolhead/BTT_EBB SB2209_CAN_RP2040_v1.0.cfg] # Do not remove this line
```
Frix-x, line 15: Can you rename the file to avoid spaces in it and prefer using "_" instead?
Also there is other BTT SB2209 files and I would prefer using the same syntax (without EBB) to keep thing aligned between all the SB2209 boards :)

So using something like : `BTT_SB2209_RP2040_v1.0.cfg` should be good :)

Frix-x (changes requested):

Frix-x:
Thanks, I'll merge for the next version :)

PR #314, opened 2023-10-07
Belt/Input Shaper improved documentation

The-Original-Reth:
I changed the format of the input shaper file, and created a belt shaper page and input shaper page. I also added more pictures.
I took the pictures and annotations from my powerpoint and incorporated them into this pull.


Frix-x:
Hello, thanks for your work on this topic. It helps me improve my documentation. I'll merge it as is and will fix the paging in order to be more "Markdown approved" if it's ok for you? :)

Also I'll add some details about the new scripts as this will be the new standard for the repository (and a lot of people are already using the v1 so I expect them to update to v2 and they will need info about them).

PR #317, opened 2023-10-15
Ability to Toggle calibration macros

NanoMetere:
my printer is already calibrated, and i don't think I need to load the macro every boot.
If I don't need to calibrate printer, it wastes ram and it increases boot time. (ram gains are small. but it can be critical if host has 1GB or 512MB ram and running a camera, and in boot time, i got approx. 1-2sec faster boot time by disabling calibration macros.)


Frix-x:
Hi, thanks for this PR!

I'm trying to understand the need behind this PR a little bit better. Indeed we discussed it with the other main contributor and most of us agree that it's a feature which adds complexity to the configuration with potential compatibility problems (if the macros are activated without an accelerometer, or if we forget them while adding an accelerometer), without adding any real benefit. Indeed, 1s-2s gain when booting Klipper is pretty insignificant compared to a print that is usually several hours...

So what is exactly the need behind this PR? Or what is the problem you are facing exactly?

NanoMetere:
yes. i agree if it just gains a small(or no) benefit and can makes a big problem for many peoples, there are not a good feature.

just realized there are no big benefits. or maybe i just want to clean the macro window but not on mainsail.

I usually print with Voron 2.4 350 with Pi4(4GB) but tested with BTT CB1(1GB)are causing a problem during print, but its running fine now with original release. I think it was a one-off issue or fixed with host system update i think?

its useless for majority so i’ll close this PR. Thanks for review!

PR #320, opened 2023-10-22
Add creality bed

Gethe:
This adds a new file to support the thermistor used by stock Creality heated beds. I also copied the existing heated_bed.cfg to the new folder as keenovo.cfg, with the old file edited to just include the copied file for backward compatibility.


Frix-x:
Hi thanks, for this PR!

It's a good idea to externalize the sensor like that and I'm wondering about doing the same for the extruders. Thanks for the idea, I'll merge this one for the next release !

PR #322, opened 2023-10-24
fix status LEDs for separate nozzle and logo LEDs

kdomanski:
The macro aways uses TRANSMIT=0 for the logo LEDs, which assumes that the subsequent call to nozzle LEDs will flush the data. That will only work if both sets of LEDs are on the same chain.

The chains can be different though - configured through `status_leds_nozzle_led_name` and `status_leds_logo_led_name`. With this change, separate logo LEDs will use TRANSMIT=1 instead.


kdomanski:
Tested on my Voron v0.2, with logo LEDs connected to the main MCU and nozzle LEDs connected to toolhead MCU.

macros/hardware_functions/status_leds.cfg
```diff
@@ -96,6 +96,9 @@ gcode:
 [gcode_macro STATUS_LEDS]
 gcode:
     {% set color = params.COLOR|default('off')|lower %}
+    {% set logo_leds_name = printer["gcode_macro _USER_VARIABLES"]["status_leds_logo_led_name"] %}
+    {% set nozzle_leds_name = printer["gcode_macro _USER_VARIABLES"]["status_leds_nozzle_led_name"] %}
+    {% set logo_transmit = 0 if (logs_leds_name == nozzle_leds_name) else 1 %}
```
Frix-x, line 101: I think there is a mistake in the variable name. Can you fix it? I'll merge it when it's done :) 
```suggestion
    {% set logo_transmit = 0 if (logo_leds_name == nozzle_leds_name) else 1 %}
```

Frix-x (changes requested):

Frix-x:
Hey, thanks for the PR, it's a great patch! I'll merge it after you fix the typo :)

macros/hardware_functions/status_leds.cfg
kdomanski: Ah, nice catch. Fixed!

Surion79:
Very nice! 👍🏻

Frix-x (approved):
Thanks, I'll merge it :)

PR #327, opened 2023-10-28
fix: module heatsoak bed should use current temperature

pvpacheco:
I've noted that the heatsoak_bed macro would always wait 8 minutes even when bed current temperature was already at the target. Maybe because at this point of the execution target is allways 0.

Changing `printer.heater_bed.target` to `printer.heater_bed.temperature` at the condition fixed that behavior.

Also fixed an inverted comment at the same file.


Frix-x (approved):
Thanks for the change! In the past it was not possible to do that directly but since the START_PRINT is not monolithic anymore, I'll merge your change that is more logical :)

PR #328, opened 2023-10-29
Create adxl345_sht_v2.x.cfg

blutack:
The SHT36_v2 differs from the SHT36_v1 and has the ADXL345 on SPI2, not SPI1. The Mellow docs are pretty confusing on this!

I'm not sure if this is the right way to go about it, or whether you would prefer to inherit the adxl345_sht.cfg and just override spi_bus?


Frix-x:
Hey, thanks for the PR!

It's ok for me like this as I do plan to work on the ADXL a little bit in the future to harmonize everything. So I'll integrate it like that for now :)

Frix-x (approved):

PR #332, opened 2023-11-06
Qgl flash fix

Surion79:
I think I found the issue with the flashing when doing QGL when G28 is not required. To fix it, the lights are only set to homing if a home is really executed.


Frix-x (approved):
Thanks for the fix. I'll merge :)

PR #333, opened 2023-11-06
added  central zhop for homing override

Surion79:
to be able to work on the zhop, I created an own PR for this.
Open for discussion

### DIFF
--- a/macros/base/homing/homing_override.cfg
+++ b/macros/base/homing/homing_override.cfg
@@ -82,6 +82,21 @@ gcode:
 
     G90
 
+    {% if ('z' not in printer.toolhead.homed_axes)  %}
+        {% if verbose %}
+            { action_respond_info("Z height unknown, performing Z-hop...") }
+        {% endif %}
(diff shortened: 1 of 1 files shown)


Surion79:
Testing this right now and the issue is a bit clear. if this ever gets called, z is fictiously homed and may cause issues, since it is homed without requesting it.

austinrdennis:
`SET_KINEMATIC_POSITION X=0 Y=0 Z=0`

I think this line here is responsible for the Z height issue you're having. You're literally telling the printer it's at position 0, 0, 0 wherever it happens to actually be. Then when you home XY only, Z thinks it's a zero and does a Z hop and thus it thinks it's at "current position + 5" on Z.

You wouldn't see it before the change because that was previously in the Z homing section and directly after it's called the printer homes every axis. I think it's in there to prevent any weird "move out of range" errors when homing all axes.

Surion79:
PR was for testing purposes with correct kinematic settings.

PR #337, opened 2023-11-09
START_PRINT macro : transmit rawparams for _MODULE_CUSTOM1/3 macros

bpontleve:
Hello,
I propose this PR, to allow access to parameters passed to PRINT_START from _MODULE_CUSTOM1/2/3 overrided macros.

Thanks to Frix_x and the team for this wonderful project.
Bruno

PS: In my case, this allows me to define a bedmesh profile name from my slicer. And call Klippain '_MODULE_BED_MESH' if not defined.

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -122,11 +122,11 @@ gcode:
         {% elif action == "extruder_preheating" %}
             _MODULE_EXTRUDER_PREHEATING
         {% elif action == "custom1" %}
-            _MODULE_CUSTOM1
+            _MODULE_CUSTOM1 {rawparams}
         {% elif action == "custom2" %}
-            _MODULE_CUSTOM2
+            _MODULE_CUSTOM2 {rawparams}
         {% elif action == "custom3" %}
-            _MODULE_CUSTOM3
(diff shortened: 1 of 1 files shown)


Surion79:
the modules are meant as defining the start sequence. never had the idea you proposing. 
Question: Do you use your custom module also in the start sequence? or is it only used as called via a parameter?

Surion79 (approved):

Frix-x:
This is a good idea, thanks for the PR!
Also, you idea about the optional bed mesh is pretty smart and I think it could be a good candidate to be added as well directly in the module bed mesh for everyone. May I ask you to add it to the same PR? :)

bpontleve:
With great pleasure, Frix_x.

Done (commit 1367666).
This is my first PR, I don't know how to add this second commit to this initial PR.

(I haven't work with the case of the non-existent bed mesh profile).

Surion79:
i thought he meant that you add rawparams to the bed mesh call i believe

Frix-x:
> i thought he meant that you add rawparams to the bed mesh call i believe

No no, @bpontleve did exactly what I understood that he made on his side. It's perfect, I'll merge it :)

PR #341, opened 2023-11-14
feat(mcu): Manta M8P V2

dhazelett:
One strange thing is that EXP header is now a ribbon, but the default config for it still defined EXP1 and EXP2 so I stuck with that. Originaly I had written it as a single, but EXP1 & EXP2 fit the default Klippain structure better.

Here is what I had written, fwiw.

```
    # EXP ribbon
    EXP1_1=<5V>   , EXP1_2=<G>  ,
    EXP1_3=PF11   , EXP1_4=PF12 ,
    EXP1_5=PF13   , EXP1_6=PF14 ,
    EXP1_7=PF15   , EXP1_8=PG0  ,
    EXP1_9=PG1    , EXP1_10=PE7 ,
    EXP1_11=<RST> , EXP1_12=PE8 ,
    EXP1_13=PE14  , EXP1_14=PE10 ,
    EXP1_15=PE11  , EXP1_16=PE15 ,
    EXP1_17=PE12  , EXP1_18=PE13
```

pinout ref https://github.com/bigtreetech/Manta-M8P/blob/master/V2.0/Hardware/BIGTREETECH%20MANTA%20M8P%20V2.0%20PinOut.png


Frix-x:
Thanks for this PR!
Actually, I try to fit the manufacturer files to what is on the board. So I think if it's a single ribbon, it should be put as a single EXP header as you've made it first.

Then can you add, the corresponding user templace file with the Klippain pins? This would make it a complete package to use the Manta M8P V2. Thank you :)

dhazelett:
@Frix-x updated, I set `LIGHT_NEOPIXEL` to `MCU_M5_STOP` like the octopus boards since v2 dropped the second RGB header.

Frix-x (approved):
Thank you for the changes. It looks good for me and I'll merge it :)

PR #342, opened 2023-11-14
Add RP2040 SPI1 USB boards

lukschwalb:
I tried to add my FYSTEC PortableInputShaper board but it uses the SPI1 of the RP2040.
This makes it neccessary to assign all SPI Pins. I guess it's quite the rabbit hole to implement all configs for every board out there but just overriding the adxl345_usb.cfg also doesn't work because it sets a value spi_bus.


Frix-x:
I'll merge it now and will add the printer.cfg entry. Thanks for this :)

PR #344, opened 2023-11-17
Add chamber heater support

mypompon:
I suggest adding the support of a chamber heater (like  https://www.teamfdm.com/uploads/monthly_2022_02/1752255662_MountedHeater.png.795b3113c8adf9e18d94891ff540a846.png).
This requires the addition of 2 pins: CHAMBER_HEATER and CHAMBER_FAN.
I've included them in the template file for Octopus (and still, there aren't enough pins available) but I don't know how to do it for other boards...


Surion79:
personally it is a pass for me, active heating is hazardeous.
Not sure about Frix wants to risk liability or anything by providing this as feature of klippain.

mypompon:
Yes, I can understand you.
Personally, I haven't had much use for this heater yet.
On the hardware side, I've included a thermal fuse in case it gets out of control.

Surion79:
what purposes does the additional chamber sensor code is needed? looking at it, i don't see the requirement. Care to elaborate? :)

mypompon:
it's because when it's a classic sensor, you get its value via : 
`printer['temperature_sensor name-of-sensor'].temperature`
but when the sensor is associated with a heater, its value is accessible via :
`printer['heater_generic name-of-sensor'].temperature`

mypompon:
To be on the safe side, I was also tempted to add a heating timeout, to turn off the heating after a given time (e.g. after 1h, even a 12h print...). 
I still have little experience with ABS, but it's mainly at the start of the print that I run into difficulties (not sticking to the bed, warping...).

Surion79:
i use nevermore as bed fans and use chamber soak functionality of klippain before printing. I set slicer to 45°C chamber temp and a 45 minute soak timer (it triggers whatever hits first). my prints are sticking goooooood

mypompon:
You mean the nevermore, located under the bed, so it takes part in transferring heat from the bed to the chamber? great usage...

Surion79:
Yes, it is an option aside the basic bed fans, but with filtering.

Frix-x:
Hello,

I've been reviewing current PRs and have concerns about this particular one. My main issue lies with the use of active heaters. They can pose significant risks if not handled correctly, and I'm keen on preventing any potential danger to users. Additionally, this project is closely associated with Voron printers (not only but it started with them), which have a clear rule against active heaters and I want to keep it as an example.

However, given that Klippain is highly customizable, you're welcome to implement this feature on your side as an override. Moreover, As Surion suggested, using a nevermore to distribute heat in the chamber, coupled with sealing any gaps, should be more than enough for all the range of materials compatible with a any hobbyist 3D printer (on my side I reach around 60-65 degree in the chamber with this configuration).

mypompon:
Hello,
Ok, no problem, I close This PR

Issue #346, opened 2023-11-18
Klipper Startup Error "variable_bed_mesh_profile" on Python2


julianschweizer:
### Klippain branch

- [X] I confirm using the main branch

### Version

4.2.0

### Describe the bug and expected behavior

Installed V4.2.0 today from scratch. 

Following Error:

```
Option 'variable_bed_mesh_profile' in section 'gcode_macro START_PRINT' is not a valid literal: unexpected EOF while parsing (<unknown>, line 0)

Once the underlying issue is corrected, use the "RESTART"
command to reload the config and restart the host software.
Printer is halted
```

Surion79:
Could you add the call in your slicer settings? Or is it crashing  when klipper is starting up?
How do you define from scratch? Can you explain which settings/files you carried over? Some settings in the files are not looking for 'from scratch'

Surion79:
somehow this is strange. the main branch of klippain has 
`variable_bed_mesh_profile =  ""`

but your config only
`variable_bed_mesh_profile = `

can you please verify? my installation works.

Surion79:
Please provide your used klipper version and the slicer section for start print. 
What you linked would mean that everyone must have this error, but i don't have it.

Surion79:
Still wondering, why you have the issue and apparently no one else. 🤔
For me the printer boots up, and we have a couple users running on dev branch to test these changes. 

But i have an idea to implement which might correct your issue also.

ndanyluk:
Just wanted to add that I'm seeing this on an upgrade from 4.1.1 to 4.2.0 as well.  This is on Python 2.7.18

ndanyluk:
FYI for those not using KIAUH, here are the steps per one of the klipper team members (I think): https://klipper.discourse.group/t/process-for-migrating-to-python3/5292/3

They seem pretty combative, but given Debian is finally moving away from packaging Python2 at all, I think Python3 may start to become the default.

ndanyluk:
@Frix-x @Surion79 just want to confirm that migrating up to Python 3 solved this for me.  Also note that you need to install numpy for Python3 as well (last post in the Klipper thread above)

Frix-x:
Ok so this confirm my hypothesis.
As Python3 is now the default, I'll not merge the PR as there is no need to absolutely keep Python2 compatibility in Klippain and we should rather focus on new features :)

PR #347, opened 2023-11-19
change string vars declaration to literals

Surion79:
seems that in some instances that empty strings in string var declarations are not recognized correctly. 
Linked to #346 

This is an untested fix, where all var declarations with strings are set as literals.

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -375,7 +375,7 @@ gcode:
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
 
     {% if bed_mesh_enabled %}
-        {% if BED_MESH_PROFILE == "" %}
+        {% if BED_MESH_PROFILE == '""'' %}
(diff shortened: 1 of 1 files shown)


Frix-x (commented):
I need to test it on my side to be sure it doesn't add any regression.

It's very strange since a lot of people are using Klippain and updated to the last version without any problem so this user may have something wrong in his configuration. I wonder if this is not a Python2 problem.

The default Klipper install is now on Python3. Also, Python2 is deprecated since January 1, 2020 so I think Python2 support in Klippain is not really something we should care about and we should instead incititate users to update their environnements.

I want to confirm this hypothesis first with this user.

Surion79:
Yes i was thinking about python as well but thought if kiauh is used, the correct version would have been selected. This is more a draft pr

Frix-x:
Ok since it was confirmed to be a Python version problem (not working under Python2), I'll close this PR in favor of Python3 that is now the default for all new install and for the other users, we can push them to update their Python environement to something more recent.

PR #348, opened 2023-11-20
Add user variable for retract length

tehniemer:
Added user variables for retract length and modified the prime line macro to alleviate an induced issue.


Surion79:
could  you describe more what your goal is in more detail?  
you removed in the prime line macro 2 important "charge up" macro lines without "optioning" them. 
You also changed in the cleaning nozzle macro 2 commands into one, not taking the different speeds into account. 
I kindly request more information about this PR. In the current state I would not approve it.

tehniemer:
I don't see the necessity of the "charge up" lines, charging can be accomplished by appropriately setting the prime line length and purge distance. 

Maybe I don't understand the reasoning behind two different speeds for the retract in the nozzle cleaning macro. Can you explain it to me?

tehniemer:
I made a few more changes for consistency across macros. I'm not sure if adding `|float` after the length variable when it's called for an extruder move is strictly necessary, but I left it in.

tehniemer:
Well, those didn't work out. I'll just stick to the original scope.

Surion79:
The different speeds to explain is something for @Frix-x . That I didn't include :)
The charge up creates a blob. The prime line is a prime line :D it doesn't need to be pretty all the way through.

Regarding the settings, i do wonder if there is a change, that they are placed in the filament section, to be set flexible by filament type.

image

macros/helpers/prime_line.cfg
```diff
@@ -80,16 +82,14 @@ gcode:
     G0 X{prime_line_x} Y{prime_line_y} F{St}
     G1 Z{line_height} F{Sz|int / 2}
 
-    # Add pressure in the nozzle
-    G92 E0
-    G1 E18 F300
-
```
Surion79, line 86: this should be either put in a different PR or created as option, so klippain users can decide if they want it or not. (default is "yes", to comply to old functionality if the variable is not defined)

Surion79 (changes requested):

macros/helpers/prime_line.cfg
tehniemer: Makes sense to add this as a user variable, I'll work on adding those options.

tehniemer:
Sorry, I'll revert the last change and create a new PR for it.

tehniemer:
I'm going to split this into two PRs, one for the user defined retract variables and another for the prime line modification. Sorry for all the hassle and back and forth.

PR #351, opened 2023-11-20
Correct board_pins to sb2209_mcu

LongDono:
Change board_pins...
From: sb2040can_mcu
To: sb2209_mcu

I originally was going to go with sb2209rp2040, but there's no need to make the alias that long.


Surion79 (approved):
i agree, the naming is matching with the BTT naming, although a bit tricky because of the two sb2209 variants

Frix-x (approved):

PR #356, opened 2023-11-22
Remove _TIP_SHAPING from End/cancel print macros

tehniemer:
I'm not sure why this is called here, but it leaves an undesirable string that can cause issues with homing Z if not removed manually.


Surion79:
please look up the files of this PR. they include other files as well.

tehniemer:
Sorry, I'm really bad at contributing with GitHub, I don't really know how to only include my changes.

Surion79:
create a branch based on klippain/dev and then add only the tip shaping changes to it and commit, publish and create PR

PR #357, opened 2023-11-22
Remove _TIP_SHAPING from End/cancel print macros

tehniemer:
I'm not sure why this is called here, but it leaves an undesirable string that can cause issues with homing Z if not removed manually.


Frix-x:
Hello,

I'd like to explain the rationale behind the TIP_SHAPING call: at the end of a print, we retract the filament above the heatbreak, this macro allow creating a clean, well-shaped tip that's ready for easy reinsertion later on.

There are two key benefits to this approach:
1. Since the filament is retracted beyond the heat-affected zone, it doesn't stick inside the nozzle. This makes filament changes easy, even when the printer is cold. Simply open the extruder latch and pull the filament out. I use this method daily.
2. It also helps in preventing filament degradation during the start_print process, which can sometimes be pretty long and involve periods with the nozzle hot without printing. By keeping the filament above the hot zone until the final purge, it avoids the risk of cooking and clogging the filament inside the nozzle.

On my side it's not a problem since even if the first homing is wrong due to this small string, It's usually cleaned during the purge or clean process above the brush before the final measurement that is done at printing temperature on my machine. But as I don't have a TAP, I agree that it could be problematic.

Does anyone else see this as a problem ? (@Surion79 @Benoitone @elpopo-eng ?)

tehniemer:
Filament retract happens after the TIP_SHAPING call at the end of the print, and this PR doesn't make any changes to retract settings.

Frix-x:
> Filament retract happens after the TIP_SHAPING call at the end of the print, and this PR doesn't make any changes to retract settings.
> 
> I know the commit history is a mess, but after the revert it's just [this one](https://github.com/Frix-x/klippain/pull/357/commits/a6b0859479adf1bd9be5feb162d7054c59cf3fb7) in this PR, or is intended to be.

Yes you're right, but the TIP_SHAPING was added on purpose to get a clean and nice tip shape in order to facilitate the reinsertion later on. But I'm fine with removing it if it cause more trouble: I can cut the tip manually before reinsertion as well.

I just need to validate that it will not cause problem to ERCF users since I know that the TIP_SHAPING part is usualkly very important for them. But maybe this is already handled correctly by HappyHare. @Benoitone @Surion79 do you have infos about this? If it's needed for you, we can also put it in a "if mmu enabled, do tip shaping" block.

Surion79:
tip_shaping is used only in the constellation with ERCF if the filament is unloaded (ERCF_UNLOAD). I think if someone uses ERCF and does not unload, i tend to think that the user doesn't want to have it reloaded. so i think it is not necessary for ERCF users
@Benoitone what is your take on this?

Frix-x:
So, I'll merge this PR now and if the TIP_SHAPING is needed for the MMU/ERCF users, then we will handle this in the HappyHare opened PR directly to restore this feature for these users.

Frix-x (approved):

agravelot:
Dragon hotend have a small lip above the heatbreak, without a proper tip shaping, it can be stuck inside. 

Thread on voron discord: 
https://discord.com/channels/460117602945990666/830395563094442044/1179962754497511454

Surion79:
> Dragon hotend have a small lip above the heatbreak, without a proper tip shaping, it can be stuck inside. 
> 
> 
> 
> Thread on voron discord: 
> 
> https://discord.com/channels/460117602945990666/830395563094442044/1179962754497511454

Better open an issue. Commenting closed PRs tend to be overseen.

PR #359, opened 2023-11-22
Add user configurable variables for retract distances in macros

tehniemer:
Added two new variables for retract, one for the cancel and end print macros and another for the nozzle cleaning macro.


tehniemer:
Breaking this into two PRs

PR #360, opened 2023-11-23
Add KNOMI support

mypompon:
I suggest you add Knomi support (tested in v1, but V2 identical).
Knomi uses 5 variables in a gcode named _KNOMI_STATUS (https://bigtreetech.github.io/docs/KNOMI2.html#adding-relevant-macros).

BTT offers a functional configuration, but overrides gcode already overridden by klippain (M109, M190, G28 (via home_override) and BED_MESH_CALIBRATE). 

However, I've added a connection with SET_HEATER_TEMPERATURE so that it also works from Mainsail/Fluidd. As SET_HEATER_TEMPERATURE is asynchronous, I've set up a timer to reset Knomi variables.

(I probably made a mistake with github, because although I had created a branch, it contained code from a previous, pending PR (Add chamber heater support). I don't know what to do if this PR is accepted and the other one isn't...)


mypompon:
I will propose a new PR without my Github mistake...

PR #361, opened 2023-11-24
Add support for Knomi

mypompon:
I suggest you add Knomi support (tested in v1, but V2 identical).
Knomi uses 5 variables in a gcode named _KNOMI_STATUS (https://bigtreetech.github.io/docs/KNOMI2.html#adding-relevant-macros).

BTT offers a functional configuration, but overrides gcode already overridden by klippain (M109, M190, G28 (via home_override) and BED_MESH_CALIBRATE).

However, I've added a connection with SET_HEATER_TEMPERATURE so that it also works from Mainsail/Fluidd. As SET_HEATER_TEMPERATURE is asynchronous, I've set up a timer to reset Knomi variables.


Frix-x (changes requested):
Hello,

Thanks for this PR!
I think this one can be merged pretty quickly since it's a simple feature.

However I have a small change request to simplify it: indeed, these additions doesn't impact the user if activated without a KNOMI connected (as it's only some variables). So I would suggest to remove the `knomi_enabled` variable and just add the macros in the default integrated macros for everyone (from the `machine.cfg` file). This way it will be always enabled in the background and if a KNOMI is connected, it will just work without having to deal with a new `printer.cfg` addition. What do you think about this?

mypompon:
Hello Frix-x,
Yes of course, I totally agree with you.
I was afraid of being too intrusive without the boolean...
I'll take care of it later today.
Thanks for your time.

mypompon:
Done.

Surion79:
Since i added a pr to knomi repo some months ago i would like to review this pr before merge @Frix-x

Surion79:
@Frix-x I honestly don't like idling delayed gcode, which is not used by 99% of the users being in the default. So I don't agree to this change. I prefer the version where you add the knomi in the printer.cfg as with every other thing. I am wondering, why this is now different for this exotic piece of equipment

Frix-x:
> @Frix-x I honestly don't like idling delayed gcode, which is not used by 99% of the users being in the default. So I don't agree to this change. I prefer the version where you add the knomi in the printer.cfg as with every other thing. I am wondering, why this is now different for this exotic piece of equipment. It is not generic, it even includes the product name.

What is the problem with this delayed gcode? It's not intrusive as it will run one or two time in the background and will just set a variable when the temperature of the heater is reached. I don't see it as a problem since it's unlikely something that will overload the gcode sequencer as there is no blocking instructions inside (ie. no movements, no wait, no M400 and only background things).

Moreover, I've chosen not to include it in the printer.cfg file because its activation doesn't affect any Klippain configurations. Keeping it enabled in the background simplifies the setup by reducing the need for additional configuration entries. My aim is to maintain simplicity for the users. Given the high volume of daily pings I receive (somedays more than 50), minimizing complexity is a priority for me, and this approach supports that goal by being user-friendly and transparent.

Surion79:
"it doesn't hurt" is the precedence for the next "it doesn't hurt". At the end it needs a rewrite and you break existing installations. so imho it is better to make it generic as always and do the diligence at the beginning to keep it clean, so additions do not begin to conflict each other or result in not being added, while it is conflicting with this. 

e.g. there is another extruder display coming up competing with knomi but uses same macros except the macros which includes the product name. So do we add it to it as well like this? What happens when they are competing? First one wins? that breaks "generic". And setting up klippain was never easy. and breaking your own designed architecture with variables in the cfg is sad.

PR #369, opened 2023-11-29
endstop_phase to enable retriving of stepper positions

Surion79:
Based on Kevins post (https://github.com/Klipper3d/klipper/pull/6193#issuecomment-1830952399), it is possible to read stepper positions by merely adding this bare tag into the printer.cfg

Then the mcu position can be retrieved by this variable
`printer.endstop_phase.last_home.stepper_x.mcu_position`

### DIFF
--- a/config/machine.cfg
+++ b/config/machine.cfg
@@ -22,6 +22,8 @@ timeout: 1800
 
 [respond]
 
+[endstop_phase]
+
 [force_move]
 enable_force_move: True
 
--- a/config/machine.cfg
+++ b/config/machine.cfg
@@ -22,7 +22,8 @@ timeout: 1800
 
 [respond]
 
-[endstop_phase]
+[endstop_phase stepper_x]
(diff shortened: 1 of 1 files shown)


Frix-x:
Oh. Never though about using this but it's clever!

I will not just directly include it for now but will wait to do some testings and validate that it will work correctly for the "automated" TEST_SPEED macro purpose (I currently don't have time to test it for now but would be happy to merge it if you add it to this PR).

Surion79:
my idea was first to add it before going for the script. but I'll take some time this weekend and put the speed test onto it. would be a possible game changer for klippain :D

Fragmon:
> my idea was first to add it before going for the script. but I'll take some time this weekend and put the speed test onto it. would be a possible game changer for klippain :D

i make a macro last year. until the new function i have to edit the stepper.py. Now its not needed. I create a PR with the macros. Feel free to improve the code :-)

PR #370, opened 2023-11-29
add_speed_calibration

Fragmon:
i create a automatic speed and acceleration test. It use the endstop phase for checking missing steps.


Surion79:
in the macro you say it is corexy or cartesian compatible. So the else paths are for corexy and cartesian for the "if"? Or is there something missing? I do wonder why corexy would name the steps A/B instead of X/Y

Fragmon:
I change this description because voron use the name A/B Drive instead X/Y:
https://docs.vorondesign.com/build/startup/

Surion79:
or for better readability it is the B/A :D (I have vorons :D)

Surion79:
if i understand it correctly, the macro is not selfiterating to higher values to determine when the steps begin to occur?

Fragmon:
what do you mean? Are you on discord? Its better to write there to improve the macro.

Surion79:
just want to understand if it is "selftuning" by calling the routine with higher values until steps were lost.

Fragmon:
No it's has no selftuning. It run until you reach the limit is reached or you loose Steps. In my github repo i have a dev Version which increase the stepper current and try this accel value again. thats very risky so i decide to delete this Feature.

Surion79:
i think what is missing is a bit like to have an end output like 

"Session finished: max accel/speed reached without  skips. Want to go higher?"
"Session interrupted: max accel/speed skipped at speed [last valid measure]. using a safety distance of 20%, the result suggest amax value of [last valid measure - 20%]"

Surion79:
testing it now on klipper 0.12.0-12

Surion79:
hmmm

`gcode_macro MAX_ACCEL_TEST:gcode': gcode.CommandError: Not enough distance on the axis for cruising. Either increase MIN_ACCEL or decrease VELOCITY.`

Fragmon:
Yes the Move length goes until the max velocity. If you accel is to low that you cant reach this speed you get the error.

Fragmon:
Fragmon. Du findest mich im voron discord.

Surion79:
I'll work on it as soon as I have the endstops reattached. Probably this weekend

EricZimmerman:
how does this compare to 

https://github.com/AndrewEllis93/Print-Tuning-Guide/blob/main/macros/TEST_SPEED.cfg

?

EricZimmerman:
and would this make more sense in Shake n Tune vs Klippain?

Fragmon:
> how does this compare to
> 
> https://github.com/AndrewEllis93/Print-Tuning-Guide/blob/main/macros/TEST_SPEED.cfg
> 
> ?

This macro have automatic step detection. Ellis macro is only a travel macro.

Frix-x:
Sorry, this was closed by mistake because I accidentely deleted the branch. @Surion79 can you push again your local "endstopphase" branch if you still have it? Sorry for that :/

Frix-x:
But I also want to point out the existence of this:  https://github.com/Anonoei/klipper_auto_speed
I tried it for the first time a couple of days ago and it was working pretty well!

Fragmon:
> But I also want to point out the existence of this: https://github.com/Anonoei/klipper_auto_speed I tried it for the first time a couple of days ago and it was working pretty well!

Wow this is very good. I like this and its more professionell then my version. You should ask him to add this into klippain.

Surion79:
> > But I also want to point out the existence of this: https://github.com/Anonoei/klipper_auto_speed I tried it for the first time a couple of days ago and it was working pretty well!
> 
> 
> 
> Wow this is very good. I like this and its more professionell then my version. You should ask him to add this into klippain.

Well, the macros are 'hidden' within the klipper extras. So not really much to do regarding integration

PR #374, opened 2023-12-04
Add configuration for Voron Decontaminator style purge bucket.

tehniemer:
Added logic to randomly select one of two user defined purge locations. Additionally clarified some verbiage for original purge bucket variable description.

### DIFF
--- a/macros/helpers/nozzle_cleaning.cfg
+++ b/macros/helpers/nozzle_cleaning.cfg
@@ -138,12 +138,21 @@ gcode:
     {% set purge_and_brush_enabled = printer["gcode_macro _USER_VARIABLES"].purge_and_brush_enabled %}
     {% set purgeclean_servo_enabled = printer["gcode_macro _USER_VARIABLES"].purgeclean_servo_enabled %}
     {% set Px, Py, Pz = printer["gcode_macro _USER_VARIABLES"].purge_bucket_xyz|map('float') %}
+    {% set decontaminator_bucket = printer["gcode_macro _USER_VARIABLES"].decontaminator_bucket|default(False) %}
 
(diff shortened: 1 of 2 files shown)


Benoitone:
Good idea 👍

Surion79:
for what would i need 2 purge buckets?

tehniemer:
It's not really two, it's a split bucket that lies on either side of the brush, so anything that falls off while scrubbing falls into the bucket. 

https://github.com/VoronDesign/VoronUsers/tree/107b1e784147c009ef9508050cbf00d441f93ca0/orphaned_mods/printer_mods/edwardyeeks/Decontaminator_Purge_Bucket_%26_Nozzle_Scrubber

Benoitone:
I have this one from the beginning  with 2 parts on my purge bucket. And with klippain only one is used. And it's true that with the "ramdom" I emptied it a little less frequently.

tehniemer:
> If it is for using another one when the other is full would not be the best option to do a random. More like variables to configure which bucket has how many possible purges within a print possible

The logic in the PR is randomly selecting one side or the other, but there is no limit to how many purges.

tehniemer:
Won't the default value of `False` here keep this from being a breaking change?

`{% set decontaminator_bucket = printer["gcode_macro _USER_VARIABLES"].decontaminator_bucket|default(False) %}`

I have a 250mm Trident so both sides of my bucket are the same size and pretty small, so this could be helpful for those with smaller beds.

Benoitone:
For me, I found the principle interesting, but indeed if it adds complexity and possible sources of error it is not really necessary... and we could forget it...

Frix-x:
I agree that depsite it being cool, it's not really needed and will close this to keep things as simple as possible

PR #377, opened 2023-12-05
Update park.cfg

smwoodward:
Added different parking positions for end vs pause


smwoodward:
I'm not really sure why it put it as 3 different PR's, but these 3 add a different end print parking position vs a pause parking.  Most of the time when the printer pauses, I want the toolhead up at the front, but when the print is over I want the head at the back to pull up the bed.  With only 1 parking position, it's always in the way at the end of the print and the motors aren't disabled so I can't move the head by hand either.

Surion79:
Shouldn't cancel_print not also result in end_park? I am not quite sure when the park is used and when end_park is used. If PARK is only used in pause, then it could be pause_park? 
Can you describe what the use case is?

Surion79:
and you used three branches (named patch-1 to patch-3) thats why it is 3 pull requests.

smwoodward:
I didn’t think about cancel also parking somewhere else and is a good idea to put it in the back. I’ll update that later today.
Sent from my iPhone

On Dec 5, 2023, at 3:31 AM, Jan-Gerrit Drexhage ***@***.***> wrote:

﻿

and you used three branches (named patch-1 to patch-3) thats why it is 3 pull requests.

smwoodward:
Ok, I've added it to the cancel print parking also to match end print.

Benoitone:
> Ok, I've added it to the cancel print parking also to match end print.

you must also put your changes in variables.cfg in your patch-1 branch.

On the other hand, could you explain to me the benefit of having several different park positions?

smwoodward:
So I did update the variables, but it did it as 3 different branches/PR’s. All 3 is the change.

The thought process for the benefit is this. When the print is paused it’s usually to work on the tool head. Feed filament, nozzle, etc. I would want the park position to be in the front for that. When the print ends, usually it’s to pull the part off the printer, which if the park position is in the front, it requires turning off motors or moving gantry to get the print off the bed. This way if it’s a pause situation the gantry can be to the front for work and if it’s a finish/cancel situation the gantry is out of the way to remove the print.

Sent from my iPhone

On Dec 9, 2023, at 1:45 AM, Benoitone ***@***.***> wrote:

﻿

Ok, I've added it to the cancel print parking also to match end print.

you must also put your changes in variables.cfg in your patch-1 branch.

On the other hand, could you explain to me the benefit of having several different park positions?

Surion79:
I'll add a review. I have an implementation improvement.

Surion79:
i think we should not have nearly two identical macros, which just have different names. I suggest making params for park and calling park without params use the variables settings and when using Pause Park, call park with the "pause park" parameters

smwoodward:
I do like that idea. Perhaps also have an XYZ parameter to use if you wanted to park it in a particular place outside of what ever variables might be saved.

Sent from my iPhone

On Dec 10, 2023, at 1:58 PM, Jan-Gerrit Drexhage ***@***.***> wrote:

﻿

i think we should not have nearly two identical macros, which just have different names. I suggest making params for park and calling park without params use the variables settings and when using Pause Park, call park with the "pause park" parameters

Surion79:
i did a PR for X and Y. i think we don't need Z (yet).

smwoodward:
I think a z hop would be good to have, if there is something hanging down that is what prompted a pause, that way it wouldn’t get dragged over the print?
Sent from my iPhone

On Dec 10, 2023, at 2:37 PM, Jan-Gerrit Drexhage ***@***.***> wrote:

﻿

i did a PR for X and Y. i think we don't need Z (yet).

Surion79:
the z hop is still used from the default value. i don't see a proper use case to have it as parameter as well. The single default value should be enough

PR #380, opened 2023-12-07
Turn on filter only if it was used during printing.

tehniemer:
I noticed my nevermore would run for 10 minutes after every print regardless of if it was used during the print, I think adding this conditional logic makes sense.

### DIFF
--- a/macros/base/end_print.cfg
+++ b/macros/base/end_print.cfg
@@ -50,7 +50,7 @@ gcode:
     # for a couple of min before stopping everything
     {% if filter_enabled and filter_speed > 0 %}
         {% set FILTER_TIME = params.FILTER_TIME|default(filter_default_time)|int %}
-        START_FILTER SPEED={filter_speed / 100}
+        START_FILTER SPEED=1
         UPDATE_DELAYED_GCODE ID=_STOP_FILTER_DELAYED DURATION={FILTER_TIME}
     {% endif %}
 
--- a/macros/base/end_print.cfg
+++ b/macros/base/end_print.cfg
@@ -12,6 +12,7 @@ gcode:
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set filter_default_time = printer["gcode_macro _USER_VARIABLES"].filter_default_time_on_end_print|default(600)|int %}
+    {% set filter_speed = printer["gcode_macro START_PRINT"].material.filter_speed|default(0)|int %}
 
     PARK
 
@@ -45,11 +46,11 @@ gcode:
     {% endif %}
 
 
-    # If a filter is connected, filter the air at full power
+    # If a filter is connected, and used during the print, continue filtering the air
     # for a couple of min before stopping everything
-    {% if filter_enabled %}
+    {% if filter_enabled and filter_speed > 0 %}
         {% set FILTER_TIME = params.FILTER_TIME|default(filter_default_time)|int %}
-        START_FILTER SPEED=1
+        START_FILTER SPEED={filter_speed / 100}
         UPDATE_DELAYED_GCODE ID=_STOP_FILTER_DELAYED DURATION={FILTER_TIME}
     {% endif %}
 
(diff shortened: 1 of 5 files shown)


macros/base/end_print.cfg
```diff
@@ -45,11 +46,11 @@ gcode:
...
 
 
-    # If a filter is connected, filter the air at full power
+    # If a filter is connected, and used during the print, continue filtering the air
     # for a couple of min before stopping everything
-    {% if filter_enabled %}
+    {% if filter_enabled and filter_speed > 0 %}
         {% set FILTER_TIME = params.FILTER_TIME|default(filter_default_time)|int %}
-        START_FILTER SPEED=1
+        START_FILTER SPEED={filter_speed / 100}
```
Frix-x, line 53: Can you keep the value of "1" here?
I prefer it to go at maximum speed at the end of the print to filter a maximum of air before opening the doors.

Frix-x (changes requested):
Thanks for this PR. It's also ok for me, good idea :)
Just a slight change asked to keep the max speed during this last cleaning process in order to filter efficiently the chamber air before opening the doors.

macros/base/end_print.cfg
tehniemer: Sure, makes sense.

Surion79:
how about adding a variable describing the functionality to be activated? So it can be either the new logic or the old

Frix-x:
> how about adding a variable describing the functionality to be activated? So it can be either the new logic or the old

No I think this small changes makes sense for everyone and do not need another new specific variable for it. I'll prefer to keep it simple and merge it like that

Frix-x (approved):

Benoitone:
doesn't work for me...

tehniemer:
Can you provide a few more details?

Benoitone:
I can't make more test now but when I print ABS (filter fan speed 30%) at the end of the print filter speed stay at 30% instead of 100%...

tehniemer:
On second look, it seems that the filter is stuck on, I think I might be improperly trying to pull a variable from `START_PRINT`, I'm guessing I have to figure out a different way to tell the `END_PRINT` macro the fan was running during the print.

{% set filter_speed = printer["gcode_macro START_PRINT"].material.filter_speed|default(0)|int %}

tehniemer:
Is there a way to poll the current `SPEED` value of a fan? This was my first choice on how to implement this because it won't require any extra variables to be passed. Maybe something like this:

```
    {% if filter_enabled and printer['fan_generic filter'].speed > 0 %}
        {% set FILTER_TIME = params.FILTER_TIME|default(filter_default_time)|int %}
        START_FILTER SPEED=1
        UPDATE_DELAYED_GCODE ID=_STOP_FILTER_DELAYED DURATION={FILTER_TIME}
    {% endif %}
```

tehniemer:
Darn, `printer['fan_generic filter'].speed` causes an error if a filter isn't enabled. Anyone else have an idea?

Benoitone:
Have a look... https://github.com/Frix-x/klippain/blob/Happy_Hare_Dev/macros/base/end_print.cfg

Benoitone:
> Have a look... https://github.com/Frix-x/klippain/blob/Happy_Hare_Dev/macros/base/end_print.cfg I didn't test but it should work

@tehniemer @Frix-x It's ok like this... I have tested... Do you want to open a PR on main branch?

Benoitone (commented):
some need modification to make it works...

tehniemer:
Have you tested that this line, `{% if printer['fan_generic filter'].speed > 0 %}`, doesn't throw an error if no filter is enabled in `printer.cfg`?

Benoitone:
> Have you tested that this line, `{% if printer['fan_generic filter'].speed > 0 %}`, doesn't throw an error if no filter is enabled in `printer.cfg`?

that's why i put `{% if filter_enabled %}` before...

tehniemer:
That still fails if no filter is enabled in `printer.cfg`. 

image

Benoitone:
> That still fails if no filter is enabled in `printer.cfg`.
> 
> image

it works for me when I comment on nevermore_filter from printer.cfg
Your error is stange if you have desactivated nevermore... maybe you have an override of it??

Benoitone:
I can put the PR if you want?

tehniemer:
> 
> it works for me when I comment on nevermore_filter from `printer.cfg` Your error is stange if you have desactivated nevermore... maybe you have an override of `[fan_generic filter]`??

You're right, I did have an override I forgot about. Go ahead if you already have a branch created for it.

Benoitone:
> > it works for me when I comment on nevermore_filter from `printer.cfg` Your error is stange if you have desactivated nevermore... maybe you have an override of `[fan_generic filter]`??
> 
> You're right, I did have an override I forgot about. Go ahead if you already have a branch created for it.

As you want... I'm ready to post a PR but if you want it's your idea...
Tell me...

tehniemer:
@Benoitone I'll do it.

PR #381, opened 2023-12-07
Add per material filament sensor switching

tehniemer:
Filament sensors, at least my version of the BTT SFS, put too much tension on flexible filaments like TPU so I've found a need to bypass it. This will allow users to enable or disable the sensor easily based on the filament type.

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -46,7 +46,7 @@ gcode:
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set firmware_retraction_enabled = printer["gcode_macro _USER_VARIABLES"].firmware_retraction_enabled %}
     {% set filter_enabled = printer["gcode_macro _USER_VARIABLES"].filter_enabled %}
-    {% set filament_sensor_enabled = printer["gcode_macro _USER_VARIABLES"].filament_sensor_enabled|default(1) %}
+    {% set filament_sensor_enabled = printer["gcode_macro _USER_VARIABLES"].filament_sensor_enabled %}
 
     {% if MATERIAL not in printer["gcode_macro _USER_VARIABLES"].material_parameters %}
         RESPOND MSG="Material '{MATERIAL}' is unknown!"
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -46,7 +46,7 @@ gcode:
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set firmware_retraction_enabled = printer["gcode_macro _USER_VARIABLES"].firmware_retraction_enabled %}
     {% set filter_enabled = printer["gcode_macro _USER_VARIABLES"].filter_enabled %}
-    {% set filament_sensor_enabled = printer["gcode_macro _USER_VARIABLES"].filament_sensor_enabled %}
+    {% set filament_sensor_enabled = printer["gcode_macro _USER_VARIABLES"].filament_sensor_enabled|default(1) %}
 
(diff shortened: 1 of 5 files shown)


macros/base/start_print.cfg
```diff
@@ -46,6 +46,7 @@ gcode:
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set firmware_retraction_enabled = printer["gcode_macro _USER_VARIABLES"].firmware_retraction_enabled %}
     {% set filter_enabled = printer["gcode_macro _USER_VARIABLES"].filter_enabled %}
+    {% set filament_sensor_enabled = printer["gcode_macro _USER_VARIABLES"].filament_sensor_enabled %}
```
Frix-x, line 49: Can you add a default value of "1" here to avoid creating a breaking change for people that don't update their material section with this new parameter?

user_templates/variables.cfg
```diff
@@ -168,7 +170,8 @@ variable_material_parameters: {
             'retract_speed': 40,
             'unretract_speed': 30,
             'filter_speed': 80,
-            'additional_z_offset': 0
+            'additional_z_offset': 0,
+            'filament_sensor': 1
```
Frix-x, line 174: Would it be possible to add your TPU section here to give it as an example with the sensor off?
Thanks :)

Frix-x (changes requested):
Thanks for this PR. I've got some small changes required before being able to merge it :)

macros/base/start_print.cfg
tehniemer: Can do.

user_templates/variables.cfg
tehniemer: Done.

tehniemer:
Something else I was thinking about, do I need to re-enable the sensor in `end_print` or does that get handled elsewhere?

Frix-x:
> Something else I was thinking about, do I need to re-enable the sensor in `end_print` or does that get handled elsewhere?



> Something else I was thinking about, do I need to re-enable the sensor in `end_print` or does that get handled elsewhere?

In fact I'm looking at it right now and there is other part of the code that deactivate and reactivate the filament sensor like in the PRIME_LINE or filament changes macros. So yes there is other parts that would need to be tweaked a little bit to avoid getting it reactivated in the middle of a print.

I think instead of relying on `filament_sensor_enabled` variable (that specify if a filament sensor is included in Klippain), we should instead focus on something like `printer['filament_motion_sensor runout_sensor'].enabled` and `printer['filament_switch_sensor runout_sensor'].enabled`

tehniemer:
I tweaked the other macros that turn the sensor on and off, but I decided to leave the way I disable it in `start_print` to be consistent with the other macros.

tehniemer:
> > Something else I was thinking about, do I need to re-enable the sensor in `end_print` or does that get handled elsewhere?
> 
> In fact I'm looking at it right now and there is other part of the code that deactivate and reactivate the filament sensor like in the PRIME_LINE or filament changes macros. So yes there is other parts that would need to be tweaked a little bit to avoid getting it reactivated in the middle of a print.

I addressed the possibility of it being re-enabled in the other macros, but do I need to modify `end_print` and `cancel_print` to re-enable it if it was disabled in `sart_print`?

Frix-x:
> I addressed the possibility of it being re-enabled in the other macros, but do I need to modify `end_print` and `cancel_print` to re-enable it if it was disabled in `sart_print`?

I think it's ok for now since when the print end it's still the same filament that is inside the machine and if you change it manually, you will likely not "print manually" without calling the START_PRINT. So it should be good to go like this :)

tehniemer:
I think there is an error with how this is used in `filament_swap.cfg`, it shouldn't adversely affect anything in those macros, but it won't function as intended. I'm going to dig into it and submit a new PR to address it. it's related to the issue that #406 aims to fix.

PR #385, opened 2023-12-07
sherpa mini config

Surion79:
used previous templates and adapted settings.
Not able to test it, since I don't own it.

### DIFF
--- a/config/hardware/extruder/sherpa_mini.cfg
+++ b/config/hardware/extruder/sherpa_mini.cfg
@@ -1,0 +1,28 @@
+[gcode_macro _USER_VARIABLES]
+variable_extruder_enabled: True
+gcode:
+
+
+[extruder]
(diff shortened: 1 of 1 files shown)


config/hardware/extruder/sherpa_mini.cfg
```diff
@@ -0,0 +1,28 @@
+[gcode_macro _USER_VARIABLES]
+variable_extruder_enabled: True
+gcode:
+
+
+[extruder]
+# Sherpa mini Gear Ratio
+# new_rd = previous_rd * mesured_distance / requested_distance
+rotation_distance: 22.4748405
```
Frix-x, line 9: Can you replace it with this one that is the "standard" one?

```suggestion
rotation_distance: 22.67895
```

Frix-x (changes requested):

PR #389, opened 2023-12-08
M8P v2.0 SPI fix proposal

Surion79:
Source

https://github.com/bigtreetech/Manta-M8P/blob/master/V2.0/Hardware/BIGTREETECH%20MANTA%20M8P%20V2.0%20PinOut.png

Issue: The Motor SPI does not have any label on the pinout. In my suggestion I renamed the SPI2 to SPI to reflect the pinout label, and the Motor as SPI2. We could also change it to MCU_M_SPI_*. But I am not sure.

What is your resolution in such matter? @Frix-x


Frix-x:
MCU_M_SPI_ is fine for me and will allow a better understanding. If there is no name on it, we can define one ourselves :)

Frix-x:
I've updated it and will merge it :)
Thanks for the draft!

PR #391, opened 2023-12-10
added parameters to park

Surion79:
I added parameters X and Y to PARK. i added two lines which can override the default settings.
This is minimal invasive and should work and can be used for multiple park positions.


Frix-x (approved):

Frix-x:
Ok this is a good idea and will allow implementing the custom park position for some macros more easily.

I'll merge this

PR #392, opened 2023-12-11
added total layer to print start

Surion79:
added the feature to have TOTAL_LAYER as parameter for Start Print


Frix-x:
I added the missing logic around the Klipper `SET_PRINT_STATS_INFO` command that is in fact why we need this ^^
I've also added some documentation to make it clear how to use it.

Frix-x (approved):

PR #396, opened 2023-12-14
Allow saving z_offset from SAVE CONFIG command

smwoodward:
With z_offset in files that are read only, if you change the z_offset and click SAVE CONFIG, klipper will not update files properly and restart with the corrected z_offset value, it requires manually putting it in the overrides.cfg.


EricZimmerman:
i would 100% discourage the use of save_config. why not simply take the value as shown in the console and manually update it in overrides.cfg?

this also breaks the "defaults" which is what is in the configs now.

i would suggest moving on from save_config and simply updating the override as you would anything else in klippain

smwoodward:
> i would 100% discourage the use of save_config. why not simply take the value as shown in the console and manually update it in overrides.cfg?
> 
> this also breaks the "defaults" which is what is in the configs now.
> 
> i would suggest moving on from save_config and simply updating the override as you would anything else in klippain

You would lean on not using save_config on everything, including PID's and bed_mesh also?  Now granted I could go in and add the defaults that are stored in the other config files for the z_offset probes in the overrides.cfg, but isn't the goal here to make things a bit more user friendly to setup?  That's great that there is some default values that are included in the configs, but those default values aren't always correct, and I would suppose that most people would use the save config to auto save those configurations and are likely not able to understand that they need to calculate that value to manually put it in the overrides.

smwoodward:
I'll give an example of the defaults not being precise.  I have used both a cnc and a printed version of tap.  The default setting for z_offset is 0.  Where the cnc tap was around -.2 on the z_offset and the printed was about -.5.  Baby stepping the offset was required for every print if I didn't press save value to calculate what the z_offset was then copying and pasting it into overrides.  

So you're correct, you and me we  understand it takes 2 seconds to copy and paste it.  This wouldn't target users like you, it would target people that don't understand.

EricZimmerman:
i would say those users wouldnt be on klippain either, but i have no data to back that up

smwoodward:
The analogous to Klippain would be RatOS, and it Klippain's readme even says it's a generic, modular, and high customizable and is designed for versatility.  This would be understood by most users that it's a way to make setting up klipper not as difficult as trying to write the entire config their selves.  I would say that ease of use would be exactly what klippain is geared towards for users that don't know the depths of klipper, and they will not understand why save_config doesn't work.  With that said, save_config only affects 3 functions.

Frix-x:
I'd like to contribute with my perspective on this topic, drawing from my previous attempts in seeking an optimal solution, although I must note there doesn't seem to be a perfect one available at the moment due to how Klipper handle this SAVE_CONFIG command.

While this command is a convenient method for saving settings like offsets, its effectiveness is limited: it only works for parameters listed in the `printer.cfg` file, not for any settings in other included configs. This limitation hinders the modular approach of Klippain, as it would require cluttering the `printer.cfg` file with numerous entries for the same components. Such an approach could lead to compatibility issues if users don't choose the correct combination of settings.

For example, in the context of PID settings, the situation is somewhat simple. Most users nowadays have both an extruder and a heated bed on their printer, so incorporating these PID values at the end of the `printer.cfg` would generally be functional. However, since Klippain allows configurations without a heated bed, including the bed PID could cause issues in these cases...

The matter of setting the z offset is way more complex. Some configurations require the probe's `z_offset` value, while others need the stepper_z `position_endstop` value. Choosing one as a default isn't feasible, as it would necessitate user intervention to select the appropriate one, as well as the correct probe type. Additionally, the beacon probe requires a different entry altogether. And these entries are not compatible between them and would lead to Klipper error if the combination is wrong in the best scenario or just incorrect settings (like z offset not applied) in the worst case due to the wrong parameter modified.

And if you apply a SAVE_CONFIG, then it will comment the line that the user has just uncommented to create this automatically generated section at the end to specifiy the same value but just differently... Given these complexities and the potential for adding confusion for novice users, I've opted not to include such dirty fix in Klippain. It's simpler to state that SAVE_CONFIG doesn't work with split configurations and use overrides instead, rather than explaining case by case which settings can be saved, which can't, and the various workarounds for saving each user needed parameters.

Finally, I just want to say that this is the official position that I have today but it doesn't means that you can't do it. Since the printer.cfg is an user file, nothing will prevent you to "override" the value you need directly at the end of the `printer.cfg` file in order to get the SAVE_CONFIG feature working for you :)

smwoodward:
The main issue that I see in trying to set it up outside of Klippain's modular files is that the entire file (in my case, Voron Tap) would need to be recreated to exclude the z_offset entry all together to allow the save config option to work as just adding it to printer.cfg doesn't fix it.  However I do see what you are saying of the complexities of the different options and that the config headings would be different for each setting, and would clutter up printer.cfg with all of those options.

I still argue that there is a novice user issue that I do think could be addressed in some fashion as it took me a bit to figure out why the save_config option wasn't working, and I'm not a novice user.  I did have enough knowledge in getting the calculated z_offset and putting it into overrides.cfg and saving the config, but most novice users would not have that level of knowledge.

smwoodward:
Here's a potential compromise it appears, looking through the configurations of probes.  While some do use position_endstop vs probe:z_virtual_endstop, all of those are under [stepper_z] and to the best of my knowledge, when you change the z offset, it only updates the z_offset under [probe].  That would be the only 2 lines that would need to be added to printer.cfg as best I can tell for all different probe options.  I think these 2 lines could be added to the probe section with some comments there explaining what will happen and it doesn't seem like it would add any more than 2 lines to the config other than the comments.

EricZimmerman:
It would tho. Take the WHOLE section and place it in printer cfg for z. Then save config will work. Don't need to mess with anything else as printer cfg overrides the reas only side. 

Simply copy ANY section you want to use save confif with to printer cfg. It has to be in printer cfg tho. 

Then save config does the rest.

Surion79:
I moved my PID settings into printer.cfg for being able to use save_config. But this only works because i know i have a heated bed, which is optional in Klippain (Frix mentioned that already).
So there is no best solution for evey way to handle it. you can only create it for yourself. (like the example of Eric)

smwoodward:
> I moved my PID settings into printer.cfg for being able to use save_config. But this only works because i know i have a heated bed, which is optional in Klippain (Frix mentioned that already). So there is no best solution for evey way to handle it. you can only create it for yourself. (like the example of Eric)

That works fine as the only reference for PID's is in the overrides.cfg file and can be moved over easily to printer.cfg.  The main issue is z_offset is listed in a read only cfg file.  Now, I understand that I can easily take (in my case) /config/hardware/probes/voron_tap.cfg and create a new file that isn't read only, remove the z_offset line out of that new file, put an include in my overrides.cfg and put 
[probe]
z_offset: 0

in my printer.cfg and I can get it to work, my argument here is that it's not novice user friendly, it even took me a moment to figure out why trying to save z_offset wasn't working, took me a bit to figure out that it was buried deeper in the config files to resolve it.  I would feel that most other people that are not as knowledgeable as we all are would have more difficulty in trying to get that to work and would not understand why it's not working.

To fix the issue so that it can work (and not confuse people who couldn't figure it out) is to remove the z_offset line under the [probe] section of all of the /config/hardware/probes config files, and move 
[probe]
z_offset:

to printer.cfg file.  I don't see where any of the z_offsets that are preconfigured would be exact and every user would need to adjust their z_offset based on their setup, which yes, can easily be added to overrides.cfg also, but I don't see how those 2 lines would clutter up printer.cfg that much.

smwoodward:
I would also argue that it would be prudent to move the control: and pid:'s for [extruder] and [heater_bed] from overrides.cfg to printer.cfg and comment them out, similar to how they are commented out in overrides.cfg.

The reason why I would say that they do need to exist in printer.cfg (besides making save_config work) is that those settings are no overrides of normal settings, they are core settings to the printer.

smwoodward:
[printer.cfg-after.txt](https://github.com/Frix-x/klippain/files/13687511/printer.cfg-after.txt)
[klippy (4).log](https://github.com/Frix-x/klippain/files/13687512/klippy.4.log)
[voron_tap.cfg.txt](https://github.com/Frix-x/klippain/files/13687513/voron_tap.cfg.txt)
[printer.cfg-before.txt](https://github.com/Frix-x/klippain/files/13687514/printer.cfg-before.txt)

Here is the proof of concept.  

First the voron_tap.cfg file is the original config file that I removed the z_offset: line from, and I saved that in my /user_mods/ folder so I could edit the file.  (I didn't necessarily go into the folder structure to edit the file with sudo to have moonraker give me issues with a dirty repo) Then I included that file in printer.cfg-before and added

[probe]
z_offset: 0
directly under the config file selection.

Attached is the logs where I homed the printer, then ran AUTO_TAP (which I've been testing a bit) which gave me a different z_offset, where I clicked save and save_config, and the printer.cfg-after is what was changed and it worked.

Surion79:
well, that would be a solution which would probably help for newbies to get klippain going. 
Especially if klipper/mainsail "insists" on using save_config. It would be a smoother user experience.

Surion79:
> Finally, I just want to say that this is the official position that I have today but it doesn't means that you can't do it. Since the printer.cfg is an user file, nothing will prevent you to "override" the value you need directly at the end of the `printer.cfg` file in order to get the SAVE_CONFIG feature working for you :)

That won't work for the z endstop, since it is already fixated in the read-only files ;) PID works already (I am using it)

smwoodward:
> > Finally, I just want to say that this is the official position that I have today but it doesn't means that you can't do it. Since the printer.cfg is an user file, nothing will prevent you to "override" the value you need directly at the end of the `printer.cfg` file in order to get the SAVE_CONFIG feature working for you :)
> 
> That won't work for the z endstop, since it is already fixated in the read-only files ;) PID works already (I am using it) The only thing making the adjustments for PID to endstop for z would mean to remove the endstop config in the defaults and move it to the overrides like the PID. So then we could copy everything over for those who know what they are doing?

does z_endstop use any z_offset settings?  I thought that was only for the probe.

Surion79:
i think i just mixed up the wording, sorry

smwoodward:
Yeah so if z_endstop doesn’t use anything for a save_config, then this wouldn’t even apply to what I’m suggesting above.

smwoodward:
Any further thought on this?

Surion79:
nope, it was tagged "wontfix". It is not being implemented.

smwoodward:
This makes no sense when I have shown that this fixes the bug that Klippain introduces with the rest of Klipper. At least with the PIDs it’s easy for users to fix that on their own, but with z_offset being buried deep in configuration files it’s not that simple. I mean even the offsets are in the overrides and commented out, just like the PIDs.

Sent from my iPhone

On Dec 18, 2023, at 11:45 AM, Jan-Gerrit Drexhage ***@***.***> wrote:

﻿

nope, it was tagged "wontfix". It is not being implemented.

Surion79:
thing is: how often do you fiddle at your z_offset? hopefully not often :D

smwoodward:
I baby step the printer quite often. As for modifying the config files to account for the baby steps, not often on one printer because I can save this configs. 

The exact same can be said for the PIDs of how often does anyone adjust those, but those can also be moved from overrides.cfg over to printer.cfg and save config will work with klippain. It is not possible to do it with z_offset unless you pull the entire probe config file and remove that one line by itself since the files are read only. 

I could easily agree with leaving the z_offset line in overrides.cfg just like the PIDs, then users can move it to printer.cfg themselves. But it needs to be removed from the other read only files.

ndanyluk:
I completely understand the issue that moving z_offset to the printer.cfg is a complex issue and would mess up Klippain's modularity.

Would a potential compromise that was stated before work?

- remove all references to position_endstop and z_offset from Klippain's read-only config
- add both settings to overrides.cfg, commented out like how we do for PID settings

This way klipper will throw an error on first run (like how it does when you don't configure PID settings), forcing a user to go and uncomment the relevant line(s).  Notes can be added to tell them which to uncomment.

Users who frequently use save config can edit from there and move to printer.cfg without issue and the core user experience isn't impacted dramatically.  As it stands, something like position_endstop for Z ALWAYS has to be calibrated by the user and forced adding these lines anyway.

Surion79:
> I completely understand the issue that moving z_offset to the printer.cfg is a complex issue and would mess up Klippain's modularity.
> 
> Would a potential compromise that was stated before work?
> 
> * remove all references to position_endstop and z_offset from Klippain's read-only config
> * add both settings to overrides.cfg, commented out like how we do for PID settings
> 
> This way klipper will throw an error on first run (like how it does when you don't configure PID settings), forcing a user to go and uncomment the relevant line(s) in overrides.cfg. Notes can be added to tell them which to uncomment.
> 
> Users who frequently use save config can edit from there and move the lines to printer.cfg without issue if they want and the core user experience isn't impacted dramatically. As it stands, something like position_endstop for Z ALWAYS has to be calibrated by the user (unless you have an infinite supply of nozzles and build plates) and forced users to add these lines anyway.

@Frix-x : For me, this looks like a very good argument. it is already done in klippain already. it is not something new. I would agree to that sort of change.

Surion79:
Why cannot be in overrides? Because it is not starting up Klipper due to missing section?

EricZimmerman:
It could but that does nothing for save_config to work which requires the section be in printer.cfg, which I thought was the point here, for save_config to work?

The real answer is for klipper to save, manage and load all save_config values from its own file (not printer.cfg) which is then applied last in the load order, but that's a different change upstream.

ndanyluk:
@EricZimmerman not sure where it was stated it can't be in overrides.cfg. All of the discussion above is about printer.cfg from what I can tell.

EricZimmerman:
You comment 5 hours ago seems contrary to this statement. It specifically mentions moving things to overrides and adding a note about people then having to manually move stuff to printer cfg as needed?

smwoodward:
> It could but that does nothing for save_config to work which requires the section be in printer.cfg, which I thought was the point here, for save_config to work?
> 
> The real answer is for klipper to save, manage and load all save_config values from its own file (not printer.cfg) which is then applied last in the load order, but that's a different change upstream.

Here's the argument that I have here in it being in overrides.cfg.  First off, everyone needs to adjust their z-offset, which is why there is already a section in the overrides.cfg for XYZ, so that people can make those adjustments in their configs.  This is the same as the PID's is that everyone needs to set those to their hardware.  The thing is, the PIDs for extruder or heated beds are not located in ANY of the read-only cfg files.  It is only located in the overrides.cfg, and if a user chooses to move the PIDs from the overrides.cfg to printer.cfg for save_config to work, then that is up to them.  However, because the z_offset section for thing is located in a read only file AND it's also included in the overrides.cfg, then a user does NOT have the option of using any of the config files that include z_offset if they would so choose to move that section over to printer.cfg.

It boils down to this.  Both z_offset and PIDs are required user adjustments.  Everyone must change their PIDs and z_offset to match their machine.  This is why it's in the overrides.  But between the two, z_offset is the ONLY setting that is included in read only files.  I say it shouldn't be in a read only file because a user must set it in the overrides.  This would not impact any user during setup, unless they forget to uncomment the [probe] and z_offset lines out of overrides.cfg, so the functionality of klippain would not change.  This would add the benefit of if a user wants to move the [probe] z_offset: settings over to printer.cfg, then it would function EXACTLY like PIDs work in klippain currently.

smwoodward:
> It cannot be in overrides, as stated, and people would have to manually move it to printer.cfg, seems like something that would confuse the type of person this change is aiming to help, which leads us back to square 1 seemingly
> 
> So this boils down to a comment to move certain config sections to printer.cfg in overrides?

The z_offset is already currently in overrides.cfg, just like the PIDs section.  So no, this would not change any functionality of Klippain as it currently sits and wouldn't be any more confusing to a user than what the PIDs in overrides.cfg currently gives.

It adds the benefit that if a user wants to move the PIDs OR the z_offset over to printer.cfg, then they could, and save_config would work.

EricZimmerman:
I have no problem with that change. Ie more stuff in overrides but that is not the same as the save_config thing being able to work out of the box. 

I agree with you on removing the stuff from RO files that needs overridden anyways and placing it in overrides.cfg. all good there. 

Most people have zero clue that the values have to be in printer cfg for save config to work so perhaps a block comment or something in overrides where all of the typical entries that save config is used for? Put the PID blocks,, probe z offset and whatever else in that same area.

smwoodward:
> I have no problem with that change. Ie more stuff in overrides but that is not the same as the save_config thing being able to work out of the box.
> 
> I agree with you on removing the stuff from RO files that needs overridden anyways and placing it in overrides.cfg. all good there.
> 
> Most people have zero clue that the values have to be in printer cfg for save config to work so perhaps a block comment or something in overrides where all of the typical entries that save config is used for? Put the PID blocks,, probe z offset and whatever else in that same area.

That wouldn't be a bad idea to have those two settings at the end of the overrides.cfg with a note stating that they could be moved over to printer.cfg for save_config to work.  

You are correct that this original PR was about trying to get save_config to work, but that is where this changed more towards being a compromise to keep klippain with it's current functionality, but allow the ability for the user to get save_config to work without having to make a copy of one of the RO files to remove the offending setting.

Surion79:
there is one thing missing in this. With missing PID you can still start Klipper (not Klippain), but i think with the missing endposition or offset, Klipper will not start at all. So there is a difference.
I don't mind moving it to overrides.cfg. if someone wants to use save_config, the user can move it manually.

PR #398, opened 2023-12-15
Create adxl345_btt_lis2dw.cfg

EricZimmerman:
Additional context or information

https://github.com/bigtreetech/LIS2DW

https://github.com/bigtreetech/LIS2DW/blob/master/BIGTREETECH%20S2DW%20V1.0%20User%20Manual_20231205.pdf

https://github.com/bigtreetech/LIS2DW/blob/master/Firmware/sample-bigtreetech-lis2dw-v1.0.cfg

### DIFF
--- a/config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
+++ b/config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
@@ -1,0 +1,21 @@
+# https://github.com/bigtreetech/LIS2DW
+[mcu adxl]
+serial: /dev/serial/by-id/xxx
+
+[lis2dw]
+cs_pin: btt_lis2dw:gpio9
+#spi_bus: spi1a
+spi_software_sclk_pin: btt_lis2dw:gpio10
+spi_software_mosi_pin: btt_lis2dw:gpio11
+spi_software_miso_pin: btt_lis2dw:gpio8
+axes_map: -y,x,-z
+
+[resonance_tester]
+probe_points: -1,-1,-1
+accel_chip: lis2dw
+
+# Include the IS calibration macros to unlock them when
+# an accelerometer is installed on the machine
+[include ../../../macros/helpers/resonance_override.cfg]
+[include ../../../macros/calibration/IS_shaper_calibrate.cfg]
+[include ../../../macros/calibration/IS_vibrations_measurement.cfg]


Surion79:
could you please update the printer.cfg with an additional entry for it?

Surion79:
and please compare to the existing cfgs. the aliases do not match. this is important :)

EricZimmerman:
like that? not sure where else to change alias. do you mean adxl345 name, or the pins on the board itself?

Surion79:
in the similar cfg files they all have `adxl:`

EricZimmerman:
yea i did change that part

config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
```diff
@@ -2,7 +2,7 @@
 [mcu adxl]
 serial: /dev/serial/by-id/xxx
 
-[lis2dw]
+[adxl345]
 cs_pin: btt_lis2dw:gpio9
 #spi_bus: spi1a
 spi_software_sclk_pin: btt_lis2dw:gpio10
```
Surion79, line 8: `adxl:` in front of the gpio

EricZimmerman: just so i am understanding, 

btt_lis2dw:gpio9

should be

adxl:gpio9

?

EricZimmerman: Any thoughts on axis map? i just took it from their example config.

Surion79:
@EricZimmerman I added some basic documentation comment lines as they exist in all other adxl files. In my opinion, this looks good.

Surion79 (approved):
OK for me.

EricZimmerman:
@Frix-x i will let you pull the trigger on this one

config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
```diff
@@ -0,0 +1,26 @@
...
+cs_pin: adxl:gpio9
+#spi_bus: spi1a
+spi_software_sclk_pin: adxl:gpio10
+spi_software_mosi_pin: adxl:gpio11
+spi_software_miso_pin: adxl:gpio8
+axes_map: -y,x,-z
+
+[resonance_tester]
+probe_points: -1,-1,-1
+accel_chip: adxl345
```
Frix-x, line 20: Same here since the name of the section is different, it need also to be changed here

```suggestion
accel_chip: lis2dw
```

config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
```diff
@@ -0,0 +1,26 @@
+# This ADXL file is dedicated to be used with ADXL boards
+# connected over USB to the pi as dedicated and standalone ADXL-MCU boards
+
+# https://github.com/bigtreetech/LIS2DW
+
+# You need to override the following to be able to set the proper serial in your overrides.cfg file
+[mcu adxl]
+serial: /dev/serial/by-id/xxx
+
+[adxl345]
```
Frix-x, line 10: This need to be changed to use the proper Klipper section for this kind of chip

```suggestion
[lis2dw]
```

config/hardware/accelerometers/adxl345_btt_lis2dw.cfg
```diff
@@ -0,0 +1,26 @@
...
+cs_pin: adxl:gpio9
+#spi_bus: spi1a
+spi_software_sclk_pin: adxl:gpio10
+spi_software_mosi_pin: adxl:gpio11
+spi_software_miso_pin: adxl:gpio8
```
Frix-x, line 15: But here is my problem... How to manage the pin name since it's not an ADXL chip anymore? I would suggest to use `accelerometer:pin_name` instead but this would need a change in all the ADXL files and is a potential breaking change if some users have overriden this in their config...

Frix-x (changes requested):
I did not answered yet since I'm not sure how to do it currently. Indeed, LIS2W is not an ADXL chip but a LIS2DW chip. According to the official manufacturer datasheet, It don't have the same comunication protocol and it is a little bit less precise and less sensitive than the standard ADXL345 (but it is also 5-10x cheaper so it explain why BTT prefer to use this one on their boards for mass production haha).

So I'm pretty sure the PR doesn't work as is and also it will need more work on the Shake&Tune side since lis2dw chips are not supported for now due to some "adxl345" hardcoded values... See this issue open there: https://github.com/Frix-x/klippain-shaketune/issues/11

But I first suggested some changes to at least make the basic accelerometer feature work in Klippain and I'll see how to update Shake&Tune to account for that later (since it's still not that a common board for now).

EricZimmerman:
so its LESS sensitive? BTT makes it sound like its the opposite!

Frix-x:
> so its LESS sensitive? BTT makes it sound like its the opposite!

That's marketing... The only advantage of the LIS2DW AFAIK is that its sampling rate is higher but that's not really important for us as the ADXL345 sampling rate is already enough for our application and probably already limited by the I2C bus speed.

JurajMik:
Maybe it can help, I have Ender 3S1 with two lis2dw probes (hotend, bed), running on biqu CB1. This is my lis2dw.cfg

`[include K-ShakeTune/*.cfg]
#[include K-ShakeTune/IS_workflow_cmd.cfg]
#[include K-ShakeTune/IS_shaper_calibrate.cfg]
#[include K-ShakeTune/IS_vibrations_measurements.cfg]

[mcu lis2dwHotendMcu]
#serial: /dev/serial/by-id/usb-Klipper_rp2040_btt_acc-if00
serial: /dev/serial/by-path/platform-5200000.usb-usb-0:1.3:1.0 

[mcu lis2dwBedMcu]
serial: /dev/serial/by-path/platform-5200000.usb-usb-0:1.4:1.0

[lis2dw lis2dwHotend]
cs_pin: lis2dwHotendMcu:gpio9
spi_software_sclk_pin: lis2dwHotendMcu:gpio10
spi_software_mosi_pin: lis2dwHotendMcu:gpio11
spi_software_miso_pin: lis2dwHotendMcu:gpio8
axes_map: -y,-z,-x #-x,-y,z

[lis2dw lis2dwBed]
cs_pin: lis2dwBedMcu:gpio9
#spi_bus: spi1a
spi_software_sclk_pin: lis2dwBedMcu:gpio10
spi_software_mosi_pin: lis2dwBedMcu:gpio11
spi_software_miso_pin: lis2dwBedMcu:gpio8
axes_map: x,y,-z

[resonance_tester]
accel_chip_x: lis2dw lis2dwHotend
accel_chip_y: lis2dw lis2dwBed
probe_points: 147, 158, 20`

Frix-x:
Also, after tests with some users equipped with the device, the LIS2DW does not seem to be really well adapted for the resonance measurements of our printers since it gets a lot of aliasing that create artifacts on the spectrogram that make it difficult to read and interpret. This noise is reproduced on the top graph and create an offset that create some low frequency signal able to move slightly the filter recommendation. It works "okay" but definitely not better than the ADXL for this purpose. I'm not saying it's a bad chip, just that it's not very well suited for this application. See this: https://github.com/Frix-x/klippain-shaketune/blob/main/docs/macros/axis_tuning.md#noisy-accelerometer

But since it's supported by Klipper, I added support for it in S&T and will merge this when the commits are fixed :)

Frix-x:
I'm closing this PR since a similar (but complete) commit was merged in c384eb263afe2627d17f9b8f233f6d26874214ad

Issue #164, opened 2023-04-26
Rainbow BARF LED Toolhead Support for StealthBurner


M3HNGRY:
Would like to request support for this if possible. 

```
# Macros for setting the status leds on the Voron StealthBurner toolhead (or for any neopixel-type leds).
#
# You will need to configure a neopixel (or other addressable led, such as dotstar). See
# https://www.klipper3d.org/Config_Reference.html#neopixel for configuration details.


#####################################
#           INSTRUCTIONS            #
#####################################
# How to use all this stuff:
#
#     1.  Copy this .cfg file into your Klipper config directory and then add [include stealthburner_leds.cfg]
#         to the top of your printer.cfg in order for register the LEDs and macros with Klipper.
#     2.  Define your LEDs by editing [neopixel sb_leds] below and entering the data pin from your control board
#         as well as the color order.
#
#           Note: RGB and RGBW are different and must be defined explicitly.  RGB and RGBW are also not able to 
#                 be mix-and-matched in the same chain. A separate data line would be needed for proper functioning.
#
#                 RGBW LEDs will have a visible yellow-ish phosphor section to the chip.  If your LEDs do not have
#                 this yellow portion, you have RGB LEDs.
#
#     3.  Save your config and restart Klipper.
#
#           Note: We set RED and BLUE to 1.0 to make it easier for users and supporters to detect 
#                 misconfigurations or miswiring. The default color format is for Neopixels with a dedicated 
#                 white LED. On startup, all three SB LEDs should light up.
#
#                 If you get random colors across your LEDs, change the color_order to GRB and restart. Then
#                 omit the W for each suggested color_order in the next paragraph.
#
#                 If you get MAGENTA, your  color order is correct. If you get CYAN, you need to use RGBW. If
#                 you get YELLOW, you need to use BRGW (note that BRG is only supported in the latest Klipper
#                 version).
#
#     4.  Once you have confirmed that the LEDs are set up correctly, you must now decide where you want 
#         these macros called up...which means adding them to your existing gcode macros.  NOTHING will happen
#         unless you add the STATUS_????? macros to your existing gcode macros.  
#
#           Example: add STATUS_LEVELING to the beginning of your QGL gcode macro, and then add STATUS_READY 
#                    to the end of it to set the logo LED and nozzle LEDs back to the `ready` state.
#
#           Example: add STATUS_CLEANING to the beginning of your nozzle-cleaning macro, and then STATUS_READY
[...]

Issue #142, opened 2023-04-11
WLED/Led_effect implementation


s0ybean:
Hi,

Just wondering if you have any ideas how an end user could implement WLED? All of the current lighting calls are baked into the non-modifiable macros and use a specific format. That makes it a bit hard to use overrides (from what I understand) since we wouldn't really need to specify rgb values or use the current format.

WLED allows us to use complex animations and offload the processing to a very inexpensive esp32. All of the setup is done on the esp32 gui.

To use it in klipper, we simply add WLED to moonraker via IP or serial, and call the specific preset that we want (which was created through the WLED interface on the esp32).

For example:

WLED_ON string=chamber preset=1

This would load any complex animations or effects we set up for preset 1 without any extra configuring on the klipper side. 

Here is the documentation:
https://moonraker.readthedocs.io/en/latest/configuration/#wled

Frix-x:
There is currently no management of WLED in Klippain. But as everything is customization, my recommendation would be to override the `STATUS_LEDS` macro by copy pasting all its content in the overrides.cfg file and adding in it some logic to also callo `WLED_ON string=xxxx preset=yyy` as you want it.

s0ybean:
thank you for the reply! :)

So, a quick way to do this would be to add this to add this in override:

```

[gcode_macro WLED_ON]
description: Turn WLED strip on using optional preset and resets led colors
gcode:
  {% set strip = params.STRIP|default("chamber")|string %}
  {% set preset = params.PRESET|default(4)|int %}

  {action_call_remote_method("set_wled_state",
                             strip=strip,
                             state=True,
                             preset=preset)}

[gcode_macro STATUS_LEDS]
    {% if printer["gcode_macro _USER_VARIABLES"].status_leds_caselight_enabled %}
        WLED_ON STRIP="chamber" PRESET={status_color[color].caselight} TRANSMIT=1
    {% endif %}

```

and change this in variables:

```
'caselight': {
            'busy': {1},
            'cleaning': {2},
            'calibrating_z': {'3},
            'heating': {4},
            'homing': {5},
            'leveling': {6},
            'meshing': {7},
            'on': {8},
            'off': {9},
            'printing': {10},
            'standby': {11},
            'error': {12},
        },
```

am i on the right path?

s0ybean:
[...]

PR #400, opened 2023-12-15
Barf stealthburner leds integration

smwoodward:
I've attempted to integrate https://github.com/julianschill/klipper-led_effect into Klippain.  I'm not 100% sure if this is formatted the way that you would want it formatted, and it certainly could use some edits to get it the way that you want it with Klippain, but this should get it started for https://github.com/Frix-x/klippain/issues/142 and https://github.com/Frix-x/klippain/issues/164


Surion79:
I am not sure that it is a good idea to have a hard dependency on another repo in the install script for a non default function.
Klippain also supports HappyHare but does not install it with klippain installation

Surion79:
Klippain includes already large portion of this PR, just in a different place. Also it feels strange to move whole sections of existing configuration away in a different file. Also I don't like opening up a whole new folder just for another repo.

For me the positive thing is that STATUS_LEDS is called in some places, which weren't used before. But keep in mind, that having something finished and started next right away with "start led" color, will result in strobe light flash, since the time between previous end color and start color of the next activity is like a couple milliseconds.

smwoodward:
> I am not sure that it is a good idea to have a hard dependency on another repo in the install script for a non default function. Klippain also supports HappyHare but does not install it with klippain installation

That's a fair point.

smwoodward:
> Klippain includes already large portion of this PR, just in a different place. Also it feels strange to move whole sections of existing configuration away in a different file. Also I don't like opening up a whole new folder just for another repo.
> 
> For me the positive thing is that STATUS_LEDS is called in some places, which weren't used before. But keep in mind, that having something finished and started next right away with "start led" color, will result in strobe light flash, since the time between previous end color and start color of the next activity is like a couple milliseconds.

The new folder I figured would be best to have all of the customizable color shows in it and keep it away from other files.  That way it couldn't get messed up by accident if it was in variables or overrides, especially since they are so large.  I also made it serveral different config files to attempt to make it where if you are also using neopixel case lights, it would give some customization.  For example I have 3 neopixel strips, BARF on stealthburner and the nozzle lights, and I use the center strip as an extension of the BARF for status indication.

I don't think I moved large portions of existing configurations out of the setups, I did add the different things for klipper led effects.  I could be wrong, but that's why I wanted to open the PR is to get some help and collaborate on it.  I know it likely needs quite a bit of work.

smwoodward:
You mentioned something about the led's strobing/flashing.  I've been using the klipper-led_effects for quite a while now and I've never noticed flashing when changing scenes.  I think it might be based on how that module handles changes.

Surion79:
If you don't mind i would create an own PR and try to integrate a bit more in klippain as it is done with other integrations.

smwoodward:
I do see how this PR has the fix for save_config for the z_offset and not just the barf integration.  I'll see if I can change the PR to fix that.

With that said, I did integrate the BARF things as much as I could tell of how klippain does it's other integrations.  If I'm missing something, can you give me a better example?

Surion79:
I would first see to have a general led_effect integration which could be enabled by uncommenting a line in the printer.cfg. Not sure if barf stealthburner is required. Issue would be that it probably collides with the other led integrations. 
But need to see how to integrate in a more generic way.
(And copying existing config files to other read only sections is bad.)

smwoodward:
> I would first see to have a general led_effect integration which could be enabled by uncommenting a line in the printer.cfg. Not sure if barf stealthburner is required. Issue would be that it probably collides with the other led integrations. But need to see how to integrate in a more generic way. (And copying existing config files to other read only sections is bad.)

The original status_led.cfg is still there, I have added the barf in addition to it.  But what did I copy from an existing config file and put it in another section?

Right now barf uses klipper-led_effect and is a separate setup compared to the original status_led setup and you wouldn't be able to use both of them at the same time, as they are different hardware required.  Currently, this wouldn't interact with other LED setups like neopixel caselights, but I am going to try and expand the klipper-led_effect to be able to handle variable caselight setups.  I am currently using this on my klippain setup in a bit of a different form.  

But there is already a general led_effect integration in the printer.cfg, this is a separate integration.

smwoodward:
Actually, looking at this, it appears that some of the ideas that I was trying to put in place didn't get pull properly.  So I'm going to close this and figure out what happened.

PR #401, opened 2023-12-16
Add support for dockable probes with separate retainer

tolson4:
Adds support for probes like the [KlackEnder probe](https://github.com/kevinakasam/KlackEnder-Probe) which use a retainer separate from the dock when detaching the probe.

### DIFF
--- a/macros/base/probing/dockable_probe.cfg
+++ b/macros/base/probing/dockable_probe.cfg
@@ -211,9 +211,8 @@ gcode:
 
     {% set probe_servo_enabled = printer["gcode_macro _USER_VARIABLES"].probe_servo_enabled %}
 
-    # Probe retainer
-    {% set probe_dock_separate_retainer = printer["gcode_macro _USER_VARIABLES"].probe_dock_separate_retainer|default(False) %}
-    {% set probe_dock_retainer_location_z = printer["gcode_macro _USER_VARIABLES"].probe_dock_retainer_location_z|default(5)|float %}
+    # Probe stow height
+    {% set probe_stow_z_height = printer["gcode_macro _USER_VARIABLES"].probe_stow_z_height|default(None) %}
 
 
     _ENTRY_POINT FUNCTION=DOCK_PROBE
@@ -262,9 +261,9 @@ gcode:
         # Move probe into dock
         _PROBE_MOVE_TO LOCATION='dock' SPEED={probe_dock_speed}
 
-        # Drop probe to retainer (if needed)
-        {% if probe_dock_separate_retainer %}
-            G0 Z{probe_dock_retainer_location_z} F{z_drop_speed}
+        # Move probe to stow z height if defined
+        {% if probe_stow_z_height is not none %}
+            G0 Z{probe_stow_z_height} F{z_drop_speed}
         {% endif %}
 
         # Get detach probe
--- a/macros/base/probing/dockable_probe.cfg
+++ b/macros/base/probing/dockable_probe.cfg
@@ -211,6 +211,10 @@ gcode:
 
     {% set probe_servo_enabled = printer["gcode_macro _USER_VARIABLES"].probe_servo_enabled %}
 
+    # Probe retainer
+    {% set probe_dock_separate_retainer = printer["gcode_macro _USER_VARIABLES"].probe_dock_separate_retainer|default(False) %}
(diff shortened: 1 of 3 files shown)


Surion79:
Looks interesting. Since I don't use docks, I can't verify it.

Frix-x:
Thanks for this PR!
Do you have a video of this? I'm not sure to understand how this kind of probe is working.

macros/base/probing/dockable_probe.cfg
```diff
@@ -211,6 +211,10 @@ gcode:
 
     {% set probe_servo_enabled = printer["gcode_macro _USER_VARIABLES"].probe_servo_enabled %}
 
+    # Probe retainer
+    {% set variable_probe_dock_separate_retainer = printer["gcode_macro _USER_VARIABLES"].probe_dock_separate_retainer %}
```
Frix-x, line 215: Can you add a default of False here?

user_templates/variables.cfg
```diff
@@ -84,6 +84,10 @@ variable_max_bed_xy: 9999, 9999
 ## Minimum safe Z height to attach/detach probe
 variable_probe_min_z_travel: 20
 
+## If probe dock has a retainer and how far to drop to before detaching 
+variable_probe_dock_separate_retainer: False
+variable_probe_dock_retainer_location_z: 20
```
Frix-x, line 89: I'm not sure if this is possible (depending of how the probe is working), but it look like that setting a value of 0 in the z value will just do nothing more than before. So these two variables may be merged into a single one that is 0 by default and you can add some value if needed.

Also just calling it "z_hop_after_probe_drop" or something like that could be a good idea to help people understand its use. But I'm unsure for now since I first need to understand how your probe is working ^^

Frix-x (changes requested):
Regarding the code you've implemented, it looks globally fine to me. I would just add some defaults to avoid a breaking change if people don't add the two new variables to their configs and maybe use only one new variable instead of two.

Let me know you thoughts about this

Frix-x:
Oh, and looking at it again, I'm wondering if this could not just be simplified by adding two new positions to the movements: "top" and "bottom" instead of the variables that you added?

tolson4:
> Do you have a video of this? I'm not sure to understand how this kind of probe is working.

Sure, here is a [video](https://youtu.be/vwmKBlRd7OA?si=wDFEmm3yE9ktvuqs&t=10) I found showing the probe being attached and detached.

> I would just add some defaults to avoid a breaking change if people don't add the two new variables to their configs and maybe use only one new variable instead of two.

This all sounds good to me. I will spend some time later today making those changes and testing.

> Oh, and looking at it again, I'm wondering if this could not just be simplified by adding two new positions to the movements: "top" and "bottom" instead of the variables that you added?

Are you referring to this? https://github.com/Frix-x/klippain/blob/main/macros/base/probing/dockable_probe.cfg#L56-L78

macros/base/probing/dockable_probe.cfg
tolson4: Sure, resolved here: 01fded42209461f531a22cd8c4bfa2eba6236b5c

macros/base/probing/dockable_probe.cfg
```diff
@@ -211,6 +211,10 @@ gcode:
 
     {% set probe_servo_enabled = printer["gcode_macro _USER_VARIABLES"].probe_servo_enabled %}
 
+    # Probe retainer
+    {% set variable_probe_dock_separate_retainer = printer["gcode_macro _USER_VARIABLES"].probe_dock_separate_retainer %}
+    {% set variable_probe_dock_retainer_location_z = printer["gcode_macro _USER_VARIABLES"].probe_dock_retainer_location_z|float %}
```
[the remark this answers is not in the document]
tolson4: So the way that this is working now, the `probe_dock_retainer_location_z` is the z value that the toolhead will drop to. I'd be worried that if this was defaulted to 0 here, there's a chance someone could inadvertently set the `probe_dock_separate_retainer` to true _without_ modifying the `retainer_location_z` value and risk damaging or colliding with their build plate.

The video linked in the other comment should give you a better idea of how the probe is docked/undocked. I'm open to suggestions if there is a different way you'd like to implement or approach this. For now I've defaulted `probe_dock_separate_retainer` to false.

tolson4:
@Frix-x This has been simplified down to a single additional variable thanks to the suggestion made by @tsk-2222 (thank you btw), and should be ready for another review whenever you get a chance.

Frix-x:
Thanks for the change, I reviewed it and it's now ok for me :)

PR #403, opened 2023-12-17
Lift Z after brush

smwoodward:
Added Z lift after brush

### DIFF
--- a/macros/helpers/nozzle_cleaning.cfg
+++ b/macros/helpers/nozzle_cleaning.cfg
@@ -49,7 +49,8 @@ gcode:
             G1 X-40 F{Sc}
             G1 X+40 F{Sc}
         {% endfor %}
-
+        #Lift Z
+        G1 Z+20 F{Sc}
         G90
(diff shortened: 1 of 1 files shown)


barecool:
may I suggest making it configurable via _USER_VARIABLES
like 
under variables.cfg add 

`variable_purge_lift_z: 5`

nozzle_cleaning.cfg
```
{% set lift_z = printer["gcode_macro _USER_VARIABLES"].purge_lift_z|default(5)|int %}
for the G1 Z{lift_z} F{Sc}

```

Surion79:
maybe it would be easy to use existing homing zhop?

Frix-x:
We also have to take care of machines where the brush is fixed to the side of the machine and the bed is moving (like on a V0): in this case the Zhop is not needed because the distance from the brush to the nozzle is not affected. I have such a machine and that's why I never implemented this zhop in the beginning.

But I'm all for it if we manage to make it a bit more intelligent :)

PR #404, opened 2023-12-17
Fixed status leds for meshing aand nozzle cleaning

smwoodward:
I noticed that the status LEDs wasn't working for bed meshing and z-tilt/qgl.


Surion79:
Thank you, please use for each functional change an own PR. This has two mixed changes (LED setting and Z movement)

smwoodward:
I don't understand what you mean. Are you saying I should do a PR for only the meshing fix, a PR for the qgl fix, and a separate PR for z-tilt?

smwoodward:
Oh wait, I see what you're saying. I must have used the wrong branch when I submitted this.

smwoodward:
> Thank you, please use for each functional change an own PR. This has two mixed changes (LED setting and Z movement)

Ok, got that fixed.

Surion79:
now i made a mess by accident. Can you please restore the nozzle_cleaning.cfg without the empty line? the original one

smwoodward:
> nozzle_cleaning.cfg

See if that fixed it.

Surion79:
Thanks! now i got it as i wanted it :D

Surion79:
i need to check the timings because of the strob flashing we had a couple weeks ago. But looks good for me

smwoodward:
I would think this would essentially work as all of the other status functions.

Surion79:
If you look at the files you added it directly above of `_CG28` which includes already status leds right at the start. which means it would flash. Could you please move your code below `_CG28`? Probably needs some changes in `_CG28` as well, but need to check that (`_CG28` has an end status led set which would flash when your code addition would be live)

smwoodward:
> If you look at the files you added it directly above of `_CG28` which includes already status leds right at the start. which means it would flash. Could you please move your code below `_CG28`? Probably needs some changes in `_CG28` as well, but need to check that (`_CG28` has an end status led set which would flash when your code addition would be live)

I see.  If you need to move them around, please do.  I put setting the status after the initial set commands like the other helpers.  I didn't think about if there was any type of loop on the homing rename and such.

Frix-x:
Hello, not sure to understand this PR since everything is already set correctly in the main calling macros and the files you modified are conditionnaly included.

Do you have a specific need or something for this? I'm not against doing this change but then we will need to clean the other parts of the code where the STATUS_LED commands are not needed anymore. For example here: https://github.com/Frix-x/klippain/blob/67980a7df9fb089a92ec2e5b4ed8e2b66e0f394b/macros/base/homing/tilting.cfg#L15-L17

This block is the one that manage the leds for QGL and Z_TILT at the same time.

Frix-x:
> > Hello, not sure to understand this PR since everything is already set correctly in the main calling macros and the files you modified are conditionnaly included.
> > Do you have a specific need or something for this? I'm not against doing this change but then we will need to clean the other parts of the code where the STATUS_LED commands are not needed anymore. For example here:
> > https://github.com/Frix-x/klippain/blob/67980a7df9fb089a92ec2e5b4ed8e2b66e0f394b/macros/base/homing/tilting.cfg#L15-L17
> > 
> > This block is the one that manage the leds for QGL and Z_TILT at the same time. There is other examples like this one.
> 
> The status animation for me isn't working for QGL/Ztilt and meshing. I'll double check and look at things, but all other status' work.

Oh, it may be because you call directly the QGL function and not the generic Klippain tilting process. But yes in this case it would be better to move the functions as you've made it indeed...

smwoodward:
> Hello, not sure to understand this PR since everything is already set correctly in the main calling macros and the files you modified are conditionnaly included.
> 
> Do you have a specific need or something for this? I'm not against doing this change but then we will need to clean the other parts of the code where the STATUS_LED commands are not needed anymore. For example here:
> 
> https://github.com/Frix-x/klippain/blob/67980a7df9fb089a92ec2e5b4ed8e2b66e0f394b/macros/base/homing/tilting.cfg#L15-L17
> 
> This block is the one that manage the leds for QGL and Z_TILT at the same time. There is other examples like this one.



> > > Hello, not sure to understand this PR since everything is already set correctly in the main calling macros and the files you modified are conditionnaly included.
> > > Do you have a specific need or something for this? I'm not against doing this change but then we will need to clean the other parts of the code where the STATUS_LED commands are not needed anymore. For example here:
> > > https://github.com/Frix-x/klippain/blob/67980a7df9fb089a92ec2e5b4ed8e2b66e0f394b/macros/base/homing/tilting.cfg#L15-L17
> > > 
> > > This block is the one that manage the leds for QGL and Z_TILT at the same time. There is other examples like this one.
> > 
> > 
> > The status animation for me isn't working for QGL/Ztilt and meshing. I'll double check and look at things, but all other status' work.
> 
> Oh, it may be because you call directly the QGL function and not the generic Klippain tilting process. But yes in this case it would be better to move the functions as you've made it indeed...

That may be what I'm experiencing also.  Now I am using klipper-led-effect with the barf examples that is mentioned here https://github.com/Frix-x/klippain/issues/164.  I am going to attempt to add this in, and I have a separate config that translates all of the klippain status calls to what klipper-led-effect uses to start and stop those animations.

smwoodward:
> Ok I've just checked and these files are ok to be modified. So we can add your changes. But can you update the calling macros to remove the original call that are not needed anymore? Also, in this case, the START_PRINT modules will also need to be updated to remove these LED settings since they will be handled in the sub macros directly.

The only thing that is in the Print_start is a BUSY status and a PRINTING status call.  

Something that I did just notice, all of the other sub macros do usually call a status_led of READY, but I did not add that to these.  Would you want that called at the end of the sub macro?  Arguably If a print is started, it should call the BUSY status and not ready, but I would assume that if it's left as READY that when the next sub macro calls the next call you wouldn't see the READY status.

EricZimmerman:
it would be nice to have all the sub calls get their own status colors/phase indicators

smwoodward:
> it would be nice to have all the sub calls get their own status colors/phase indicators

For the most part, they already do.

Surion79:
it is important to understand, that this PR is re-introducing strobe flashing on LED status calls as it is integrated in this PR here, @Frix-x .
With these changes the leds are turned to the color and will instantaneously reflash to a different color since _CG28 will always call a Status LED every time, either when it runs G28 or even when not. 
It would need to do some code movement and probably a change in _CG28 so the flashing is not reintroduced

Frix-x:
I'll close this one in favor of #412 that is a big rework of the LED system and that contain the same feature. So this will get merged soon :)

PR #405, opened 2023-12-18
Update orbiter2.0.cfg

EricZimmerman:
Add comment about required current. At the default of 0.45 on the Orbiter 2 motor, very poor performance will occur including significant underextrusion on infill, etc.

### DIFF
--- a/config/hardware/extruder/orbiter2.0.cfg
+++ b/config/hardware/extruder/orbiter2.0.cfg
@@ -22,6 +22,9 @@ min_extrude_temp: 172
 pressure_advance: 0.0475
 pressure_advance_smooth_time: 0.040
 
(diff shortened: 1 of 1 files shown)


Frix-x:
Thanks, I see no harm to add this comment.
Also FYI, this currents settings are something pretty problematic since everyone would need to tweak them and I'm currently thinking about adding a variable to set them at Klipper startup and let some generic values like 0A current in the config file (or something very low like 0.1A if 0 is not possible). But it will add a breaking change so unsure about how to do it cleanly for now...

Frix-x (approved):

EricZimmerman:
i like that idea. perhaps a section in overrides for the common printers where they have to be set.

PR #406, opened 2023-12-18
Fix error with turning on filter only if it was used during printing.

tehniemer:
Fixed issue discovered after merging https://github.com/Frix-x/klippain/pull/380

### DIFF
--- a/macros/base/cancel_print.cfg
+++ b/macros/base/cancel_print.cfg
@@ -10,6 +10,7 @@ gcode:
     {% set light_enabled = printer["gcode_macro _USER_VARIABLES"].light_enabled %}
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
+    {% set filter_default_time = printer["gcode_macro _USER_VARIABLES"].filter_default_time_on_end_print|default(600)|int %}
 
     PARK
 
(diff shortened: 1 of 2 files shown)


Surion79:
does this work?

tehniemer:
Yes, @Benoitone and I have tested it.

tehniemer:
I also added this to the `cancel_print` macro. I don't know why it wouldn't be desired there as well.

tehniemer:
OMG, I don't know why my main branch keeps merging into these PR branches and screwing up the history.

Surion79:
I would like to check this. The coding seems not optimal for me.
Especially that it is fixed on a hard coded fan name in a read only macro. That is not generic

Benoitone:
> I would like to check this. The coding seems not optimal for me.
> Especially that it is fixed on a hard coded fan name in a read only macro. That is not generic

I still think that this is the easiest way to stop the nevermore which is also hard coded in a read-only file?
Maybe you have a better idea?

Surion79:
Klippain is generic and has multiple fan options. I need to dig in, and i think there is a more genetic solution.

Frix-x:
> I would like to check this. The coding seems not optimal for me. Especially that it is fixed on a hard coded fan name in a read only macro. That is not generic

I agree with Benoit on this point since it's also us that define this name in read-only files so there is unlikely someone that would change it (and if it does, every SET_FILTER macro will stop to work also... so it would be to define his own thing).
I'll merge it like this as it's ok for me :)

Frix-x (approved):

tehniemer:
> Klippain is generic and has multiple fan options. I need to dig in, and i think there is a more genetic solution.

I think this is likely the most generic way to execute this. It should even catch the case where a user manually turns the filter on during a print.

PR #407, opened 2023-12-19
Update start_print.cfg with FORCE_MESH

0rinsb3lt:
Added FORCE_MESH as a parameter to ADAPTIVE_BED_MESH

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -34,7 +34,7 @@ gcode:
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=material VALUE='"{MATERIAL}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=fl_size VALUE='"{FL_SIZE}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_mesh_profile VALUE='"{BED_MESH_PROFILE}"'
-    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=force_mesh VALUE='"{FORCE_MESH}"'
+    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=force_mesh VALUE={FORCE_MESH}
 
     # Get all the config options and configurations for this macro
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -22,6 +22,7 @@ gcode:
     {% set MATERIAL = params.MATERIAL|default(printer["gcode_macro _USER_VARIABLES"].print_default_material)|string %} # Material type set in the slicer
     {% set FL_SIZE = params.SIZE|default("0_0_0_0")|string %} # Get bounding box of the first layer for the adaptive bed mesh
     {% set BED_MESH_PROFILE = params.MESH|default("")|string %} # Bed mesh profile to load
+    {% set FORCE_MESH = params.FORCE_MESH|default(0)|int %}
 
     # Set the variables to be used in all the modules based on the slicer parameters
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_temp VALUE={BED_TEMP}
@@ -33,6 +34,7 @@ gcode:
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=material VALUE='"{MATERIAL}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=fl_size VALUE='"{FL_SIZE}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_mesh_profile VALUE='"{BED_MESH_PROFILE}"'
+    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=force_mesh VALUE='"{FORCE_MESH}"'
 
     # Get all the config options and configurations for this macro
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
@@ -369,8 +371,8 @@ gcode:
     # ----- BED MESH -------------------------------------------
     {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
     {% set BED_MESH_PROFILE = printer["gcode_macro START_PRINT"].bed_mesh_profile %}
-
-    {% set force_mesh = printer["gcode_macro _USER_VARIABLES"].force_mesh %}
+    {% set force_mesh = printer["gcode_macro START_PRINT"].force_mesh %}
+    
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -386,7 +386,7 @@ gcode:
                 RESPOND MSG="Bed mesh measurement..."
             {% endif %}
 
-            ADAPTIVE_BED_MESH SIZE={FL_SIZE} FORCE_MESH={force_mesh}
+            ADAPTIVE_BED_MESH SIZE={FL_SIZE} FORCE_MESH={FORCE_MESH}
         {% else %}
             {% if verbose %}
                 RESPOND MSG="Load bed mesh profile : {BED_MESH_PROFILE}"
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -370,6 +370,7 @@ gcode:
     {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
     {% set BED_MESH_PROFILE = printer["gcode_macro START_PRINT"].bed_mesh_profile %}
 
+    {% set force_mesh = printer["gcode_macro _USER_VARIABLES"].force_mesh %}
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
@@ -383,7 +384,7 @@ gcode:
                 RESPOND MSG="Bed mesh measurement..."
             {% endif %}
 
-            ADAPTIVE_BED_MESH SIZE={FL_SIZE}
+            ADAPTIVE_BED_MESH SIZE={FL_SIZE} FORCE_MESH={force_mesh}
         {% else %}
             {% if verbose %}
                 RESPOND MSG="Load bed mesh profile : {BED_MESH_PROFILE}"
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -371,7 +371,7 @@ gcode:
     # ----- BED MESH -------------------------------------------
     {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
     {% set BED_MESH_PROFILE = printer["gcode_macro START_PRINT"].bed_mesh_profile %}
-    {% set force_mesh = printer["gcode_macro START_PRINT"].force_mesh %}
+    {% set FORCE_MESH = printer["gcode_macro START_PRINT"].force_mesh %}
     
     {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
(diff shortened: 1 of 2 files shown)


Surion79:
Can you elaborate the use case? Actually it is missing something. 1. I don't see the variable in variables.cfg. 
2. is it an option for start_print as parameter within the slicer? That would be missing too

Benoitone:
> Can you elaborate the use case? Actually it is missing something. 1. I don't see the variable in variables.cfg. 2. is it an option for start_print as parameter within the slicer? That would be missing too

I think he must add {% set force_mesh = params.FORCE_MESH|default(False) %} in the param from the slicer?

Surion79:
He remarks an unknownnuser variable. So nut sure of the integration. It would mean defaulting to the variable and not to a hard coded default.

Benoitone:
?

{% set force_mesh = params.FORCE_MESH|default(False) %} in the param from the slicer?
Is just set to a default variable?

0rinsb3lt:
You are correct I didn't put it into variables.cfg. This PR is to give the user the option to force a bed mesh.
Most of the code was already there it just defaulted to False and never actually read from the variables.cfg file (User variables) or gave the end user a way to enable it. 

I will update with a defaulted variables_force_mesh: False in variables.cfg

Surion79:
i would also like to have it as parameter in start_print available, if we are already integrating it

0rinsb3lt:
> i would also like to have it as parameter in start_print available, if we are already integrating it

So that you could enable from the slicer?

Surion79:
yes. you can look the macro with PARAMS. and adapt it accordingly for bed mesh. I expect some integration since the bed mesh module which is a starting module was modified.

Benoitone:
I don't really see the real point of adding another user variable with all the problems that can cause (when upgrading Klippain) for a function that has limited interest. Why make a bed mesh when the printed part is very small?
So in my opinion it could remain something marginal that the user would be able to pass into their slicer start g_code and which would be used here https://github.com/Frix-x/klippain/blob/a20763997ef134cb24b401cb071eac245126f724/ macros/calibration/adaptive_bed_mesh.cfg#L125-L126.

And in this case I don't see how replacing `{% set force_mesh = printer["gcode_macro _USER_VARIABLES"].force_mesh %}` by `{% set force_mesh = params.FORCE_MESH|default(False) %}` could cause a problem.

Finally, in the event that you absolutely want to add the new user variable entry in `variables.cfg` the description can easily lead to confusion: the way it is worded can make you think that there will be no bed mesh if you leave it on False...

Surion79:
the whole thing makes no sense since the addition is in the bed_mesh_module within the start print sequence. you either need to pass it through start_print params or through variables. if neither is done, the addition is plain never in use. Please look at the change. it is NOT done in the START_PRINT macro.

Surion79:
and i agree, the var wording creates more issues than it solves

macros/base/start_print.cfg
```diff
@@ -370,6 +370,7 @@ gcode:
     {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
     {% set BED_MESH_PROFILE = printer["gcode_macro START_PRINT"].bed_mesh_profile %}
 
+    {% set force_mesh = printer["gcode_macro _USER_VARIABLES"].force_mesh %}
```
Frix-x, line 373: Can you remove the variable and use this line instead?
```suggestion
    {% set force_mesh = params.FORCE_MESH|default(0) %}
```

user_templates/variables.cfg
```diff
@@ -37,6 +37,9 @@ variable_zendstop_position: -1, -1 # change it ONLY if not using the auto z cali
 ## Force always a full homing and QGL/Z_TILT during the START_PRINT macro
 variable_force_homing_in_start_print: False
 
+## Always force an adaptive bed mesh during the START_PRINT macro
+variable_force_mesh: False
+
```
Frix-x, line 42: Remove this as it's not needed for this and would add more confusion since it would need a restart of Klipper each time you want to change it and it's not really flexible to configure it per print

Frix-x (changes requested):
Yes I think this doesn't need a variable but rather a START_PRINT param as it is likely to be used based on the parts to be printed directly from the slicer profile and setting a variable for that is not a good usecase.

Also please note that since this parameter is originally designed to be used from the slicer (even if in Klippain it was a little bit forgotten), so it's not using a Jinja True/False value but a 0/1 instead.

I added some suggestions to make it ready for a merge :)

macros/base/start_print.cfg
Benoitone: > Can you remove the variable and use this line instead?
> 
> ```suggestion
> 
>     {% set force_mesh = params.FORCE_MESH|default(0) %}
> 
> ```
> 
> 

I think the params must be parse in this section of print_start and call after in bed mesh module?

    # Get all the parameters passed from the slicer

Surion79: why not using the implementation style in start print like it is done previously all the time? the bed mesh module is never called directly outside start print

Surion79:
i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding

Benoitone:
> i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding

It seems perfect for me...
@Surion79 @Frix-x: Just for my knowledge... is there a difference if we use `False` instead of `0` as the default value?

Surion79:
> > i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding
> 
> It seems perfect for me... @Surion79 @Frix-x : Just for my knowledge... is there a difference if we use `False` instead of `0` as the default value in this case?

i made it int, 0/1 as frix stated it in an above comment

Benoitone:
> > > i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding
> 
> > 
> 
> > It seems perfect for me... @Surion79 @Frix-x : Just for my knowledge... is there a difference if we use `False` instead of `0` as the default value in this case?
> 
> 
> 
> i made it int, 0/1 as frix stated it in an above comment

Because in https://github.com/Frix-x/klippain/blob/a20763997ef134cb24b401cb071eac245126f724/macros/calibration/adaptive_bed_mesh.cfg#L67 it seems to be a boolean...

Benoitone:
And maybe take the opportunity to include this function in configuration.md... https://github.com/Frix-x/klippain/blob/e1a6cba3f009e8248b473754599ea1ad2f7bf8c1/docs/configuration.md

Surion79:
> > > > i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding
> > 
> > 
> > > 
> > 
> > 
> > > It seems perfect for me... @Surion79 @Frix-x : Just for my knowledge... is there a difference if we use `False` instead of `0` as the default value in this case?
> > 
> > 
> > i made it int, 0/1 as frix stated it in an above comment
> 
> Because in
> 
> https://github.com/Frix-x/klippain/blob/a20763997ef134cb24b401cb071eac245126f724/macros/calibration/adaptive_bed_mesh.cfg#L67
> 
> it seems to be a boolean...

could you please comment the line directly in the PR? I don't see it in the current version of the PR

Benoitone:
> > > > > i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > It seems perfect for me... @Surion79 @Frix-x : Just for my knowledge... is there a difference if we use `False` instead of `0` as the default value in this case?
> > > 
> > > 
> > > i made it int, 0/1 as frix stated it in an above comment
> > 
> > 
> > Because in
> > https://github.com/Frix-x/klippain/blob/a20763997ef134cb24b401cb071eac245126f724/macros/calibration/adaptive_bed_mesh.cfg#L67
> > 
> > it seems to be a boolean...
> 
> could you please comment the line directly in the PR? I don't see it in the current version of the PR

I don't see how to do this on mobile app...
it's in macros/calibration/adaptive_bed_mesh.cfg -> Ligne67

Surion79:
oooh, you mean outside this PR. Silly me :D

@Frix-x regarding your statement of force_mesh being 0/1 and in bed mesh it was programmed true/false. Which way to go? Stay int, go boolean or switch from int to bool?

macros/base/start_print.cfg
```diff
@@ -33,6 +34,7 @@ gcode:
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=material VALUE='"{MATERIAL}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=fl_size VALUE='"{FL_SIZE}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_mesh_profile VALUE='"{BED_MESH_PROFILE}"'
+    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=force_mesh VALUE='"{FORCE_MESH}"'
```
Frix-x, line 37: In both case (integer or boolean use), this needs to be changed here to avoid having a "False" string evaluated as True due to how Jinja read strings.

```suggestion
    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=force_mesh VALUE={FORCE_MESH}
```

macros/base/start_print.cfg
```diff
@@ -369,7 +371,8 @@ gcode:
     # ----- BED MESH -------------------------------------------
     {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
     {% set BED_MESH_PROFILE = printer["gcode_macro START_PRINT"].bed_mesh_profile %}
-
+    {% set force_mesh = printer["gcode_macro START_PRINT"].force_mesh %}
```
Frix-x, line 374: Please use uppercase for the var name here to keep it consistent:

```suggestion
    {% set FORCE_MESH = printer["gcode_macro START_PRINT"].force_mesh %}
```

macros/base/start_print.cfg
```diff
@@ -383,7 +386,7 @@ gcode:
                 RESPOND MSG="Bed mesh measurement..."
             {% endif %}
 
-            ADAPTIVE_BED_MESH SIZE={FL_SIZE}
+            ADAPTIVE_BED_MESH SIZE={FL_SIZE} FORCE_MESH={force_mesh}
```
Frix-x, line 389: And same here

```suggestion
            ADAPTIVE_BED_MESH SIZE={FL_SIZE} FORCE_MESH={FORCE_MESH}
```

Frix-x (changes requested):
> i took the liberty to correct it directly. Should be fine now, since it is 100% matching Module coding

Be careful by doing this, I usually avoid it because it will commit and push directly on the user repo without its consentement and that could lead to troubles especially if it's pushing directly on his main branch like for this PR (ie. worst case is directly breaking his machine by pushing some bad code). Also I like to keep the commits authered by each users as possible as it's good to keep the authorship/copyright for them but it's also easier to find out who have made what on Klippain for us.

> @Frix-x regarding your statement of force_mesh being 0/1 and in bed mesh it was programmed true/false. Which way to go? Stay int, go boolean or switch from int to bool?

Yes, I will need to test it. In Jinja, when a string like "False" is used, it can be very confusing because, as a string, it is non-null and thus evaluates as True in a boolean context. So I know that I got a lot of trouble with this in the past and remember that to make it work, I had to use `FORCE_MESH=1` on my Prusa. But this need to be tested to be sure.

Finally I added two small change request:
1. First one is a small error on the same subject that will make it not work so it needs to be changed if it's really an integer that needs to be used
2. Second one is just to keep it well formated and coding style consistent (uppercase wanted)

Surion79:
understood and done. usually I don't do that :)

macros/base/start_print.cfg
```diff
@@ -22,6 +22,7 @@ gcode:
     {% set MATERIAL = params.MATERIAL|default(printer["gcode_macro _USER_VARIABLES"].print_default_material)|string %} # Material type set in the slicer
     {% set FL_SIZE = params.SIZE|default("0_0_0_0")|string %} # Get bounding box of the first layer for the adaptive bed mesh
     {% set BED_MESH_PROFILE = params.MESH|default("")|string %} # Bed mesh profile to load
+    {% set FORCE_MESH = params.FORCE_MESH|default(0)|int %}
```
Benoitone, line 25: I don't understant why int is better than boolean in this case?
when user pass the start_print parameter: FORCE_MESH=1 is interpreted as True, and with FORCE_MESH=0 is False??

```suggestion
    {% set force_mesh = params.FORCE_MESH|default(False) %}
```

Frix-x: Yes this should be the case, but as I said, I need to test it properly since Jinja and boolean is pretty finicky.

Here is why I prefer 0/1 from the slicer:
1. If you put FORCE_MESH=False -> this should result in False so it's ok
2. If you put FORCE_MESH="False" -> this will result is True since "False" is a non-null string so it will be very confusing and I'm not sure how the slicer will create the Gcode.

But again I know that it's finicky and that I already battled with this in the past, just not sure if it's exactly this case or if it's still the case.

Frix-x:
Ok, so now the PR is ready to be merged. I just want to test it thoroughly to make sure it's working before I merge it. So please give me some time to do that (currently batch printing Christmas decorations haha)

mjonuschat:
Full fix implemented in #814

PR #411, opened 2023-12-20
Make filament sensor status checks universal

tehniemer:
Change the way the status of the filament sensor is checked so it won't be re-enabled if it was previously disabled by a macro or manually. Also add code to re-enable the sensor in the end_print and cancel_print macros.

After giving some thought about [this](https://github.com/Frix-x/klippain/pull/381#issuecomment-1853590423) comment I agree that it's a better way to universally check the status of the filament sensor prior to making any change to its state.

I've tested this in the purge routine and if the sensor was disabled before the macro was called it remains disabled after completion, and vice-versa.


Frix-x (approved):
This looks to be pretty good for me. It's exactly what I wanted to do with your first PR so this is pretty aligned with my initial idea :)

I'll merge, thanks for the enhancement of your work!

tehniemer:
Glad I can give something back to this great project.

PR #418, opened 2023-12-21
chore(mcu): properly label CAN H/L

dhazelett:
TX and RX are not helpful, looking at the back of the m8p v2 shows that RX is high, and TX is low.

While these are not actually referenced in the configs, it may still be very helpful for some as the documentation for this board can be a bit confusing.

image

IMG_8251


EricZimmerman:
Something like this makes things a lot easier imo

https://github.com/EricZimmerman/VoronTools/blob/main/img/ebb/EBBCANSetup.png

Vendor schematics are terrible for showing which ports are which.

dhazelett:
which is made even more confusing for the manta, it doesn't require a u2c or anything. I wish the vendors would put out decent docs for CAN 😂

Frix-x (approved):

PR #424, opened 2024-01-01
check if heatsoak is configured to verbose message

Surion79:
I added a check if heatsoak bed is enabled (value > 0) so it does not write misunderstandable output in console.


Frix-x (approved):
Good idea, thanks

PR #425, opened 2024-01-01
correct overriders misspelling

Surion79:
added this PR for misspelling overides as states by @EricZimmerman .
I made a replace in files to change alle misspellings and renamed the overide.cfg file correctly.


EricZimmerman (approved):
seems good to me!

Surion79:
Since it is just a string replace within the ready-only section, i think the impact should be non-existant to minimal (for those adding these specific folder files in a different own file)

Frix-x (approved):
This should be ok! I'll merge, thanks for the catch :)

PR #371, opened 2023-11-30
HappyHare v2.4.1 Dev support

Benoitone:
To align to develop branch before possibility to correctly merge.
Easiest than https://github.com/Frix-x/klippain/pull/368

PR #426, opened 2024-01-02
added spoolman support

Surion79:
I added spoolman integration to klippain.

Source: https://github.com/Donkie/Spoolman

I tried using the configuration with the provided default settings (in my case i went with docker installation on the pi) It works with localhost to point to itself in browsing the spoolman webinterface and the integration in mainsail.

### DIFF
--- a/moonraker/spoolman.conf
+++ b/moonraker/spoolman.conf
@@ -1,0 +1,8 @@
+# this is a default configuration pointing the default installation on the printer itself
+# you can set an own config by excluding this file, copy its content in moonraker.conf and change the values
+[spoolman]
+server: http://localhost:7912
+#   URL to the Spoolman instance. This parameter must be provided.
+sync_rate: 5
+#   The interval, in seconds, between sync requests with the
+#   Spoolman server.  The default is 5.
--- a/user_templates/moonraker.conf
+++ b/user_templates/moonraker.conf
@@ -31,6 +31,9 @@
 # [include moonraker/z_calibration.conf]
 # -----------------------------------------------------------------
 
+##### Spoolman configuration file  -----------------
+# [include moonraker/spoolman.conf]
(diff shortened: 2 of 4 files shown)


Benoitone:
I also added it in the Happy Hare branch :)

Surion79:
it would be easier, if you check this, instead of me finding in the dozens of modified files of your PR where you integrated spoolman :)
(and it wouild be helpful to have one function per PR and don't mix them together)

Surion79:
i think you are missing the moonraker.conf include for spoolman to work.
And i don't know if it is a good idea to create files which include only a single include to a different file.

roykrikke:
+1

moonraker/spoolman.conf
```diff
@@ -0,0 +1,8 @@
+# this is a default configuration pointing the default installation on the printer itself
+# you can set an own config by excluding this file, copy its content in moonraker.conf and change the values
+[spoolman]
+server: http://localhost:7912
```
Benoitone, line 4: for me I must use ip instead of localhost to work properly :(

user_templates/moonraker.conf
```diff
@@ -31,6 +31,9 @@
 # [include moonraker/z_calibration.conf]
 # -----------------------------------------------------------------
 
+##### Spoolman configuration file  -----------------
+# [include moonraker/spoolman.conf]
+# -----------------------------------------------------------------
```
Benoitone, line 36:
```suggestion
##### Spoolman configuration file  -----------------
# [include moonraker/spoolman.conf]
#[spoolman]
#server: http://SpoolmanServerIP:7912
# -----------------------------------------------------------------
```

Frix-x:
I'll close this PR since everything is now properly managed in the MMU management PR #371

PR #428, opened 2024-01-04
HH v2 change reference

Surion79:
This is only used as reference for the HH v2 branch and is not finished. it is only a reference for @Benoitone 

Need to test some overrides to make the code more readable


Surion79:
since HHv2 branch is now about ready to be released, this draft for reference is no longer required

PR #431, opened 2024-01-11
add retract in park macro

Benoitone:
In curent park macro the retraction of filament is only possible when firmware_retraction is included in printer.cfg. Otherwise there is no retraction.
This PR try to add some filament retraction in case of firmware retraction isn't enabeled in printer.cfg by using the curent material parse by start_print or the default (E) value already existing in park macro.


Surion79:
looks good. 
But: Do we have any checks to verify that the retraction config per filament has a valid value and not something like negative?

Surion79:
i think we can close this PR since you already added it to the mmu pr?

Frix-x:
Yes, I'll close it for now and if needed we will fix the thing in the develop branch when the MMU PR is merged :)

PR #439, opened 2024-01-20
BTT S2DW V1.0 and similar RP2040 + lis2dw boards

gralves2:
This pull request adds support for the BTT S2DW V1.0 and similar boards where a lis2dw accelerometer is associated to an RP2040 by SPI1.


Frix-x (changes requested):
Thanks for the PR. This is almost ok for me, just a little change to the filename.

Frix-x (approved):
This is ok for me. I'll merge, thank you!

PR #442, opened 2024-01-24
nevermore-controller for StealthMax and Nevermore Max

qratz:
add configuration for nevermore controller from https://github.com/Sa…naaHamel/nevermore-controller as used in Nevermore Max and Nevermore StealthMax

### DIFF
--- a/config/hardware/filters/nevermore_controller_filter.cfg
+++ b/config/hardware/filters/nevermore_controller_filter.cfg
@@ -1,0 +1,112 @@
+# Recirculating carbon filter (example: Nevermove v5, etc...)
+[gcode_macro _USER_VARIABLES]
+variable_filter_enabled: True
+variable_filter_name: "nevermore_fan"
+variable_filter_nevermore_contoller: True
+gcode:
+
+# Also include directly the filter control macros from here
+[include ../../../macros/hardware_functions/filter.cfg]
+
+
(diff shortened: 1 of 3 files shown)


Surion79:
I am not sure if it is ok to add config files which are basically completely commented out. usually including files have some sort of basic functionality which can be used out of the box.
In your integration, if i uncomment the file for the nevermore, the end_print will always trigger the nevermore command instead of the correct SPEED=0 command for the other nevermore filter. 
This stage feels a bit more like being implemented in override for the people who uses it, since it needs to be fully overridden anyway.

qratz:
I think you are right. The nevermore-controller essentially acts autonomously, which makes it a very nice solution. So just adding the [nevermore] section in overrides, enabling filter and setting the correct filtername in variables.cfg is all that is needed to enable the filter with the expected behaviour.

Frix-x:
Thanks for the discussion about this!
So I think for now nothing more needs to be added to Klippain? If this is the case, I'll close the PR :)

qratz:
Yes, the PR can be closed. Maybe it someone could add information on how to use nevermore-controller in one of the MD-Files.

Frix-x:
> Maybe it someone could add information on how to use nevermore-controller in one of the MD-Files.

Yes this is a good idea. If you want to open a PR with some documentation, I'll be happy to merge it :)

PR #448, opened 2024-02-01
Status leds on startup fix

Benoitone:


macros/miscs/startup.cfg
```diff
@@ -49,7 +52,7 @@ gcode:
 
 [gcode_macro _INIT_LEDS]
 gcode:
-    {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
+    {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled or printer["gcode_macro _USER_VARIABLES"].status_leds_caselight_enabled or printer["gcode_macro _USER_VARIABLES"].light_enabled %}
```
Frix-x, line 55: `or printer["gcode_macro _USER_VARIABLES"].light_enabled` need to be removed since there is no STATUS_LEDS macro with the white lights

Frix-x (changes requested):

macros/miscs/startup.cfg
Frix-x, line 55:
```suggestion
    {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled or printer["gcode_macro _USER_VARIABLES"].status_leds_caselight_enabled %}
```

PR #450, opened 2024-02-02
Add BLTouch support

nmsmith22389:
Added support for BLTouch probes as a virtual endstop.

### DIFF
--- a/config/hardware/probes/bltouch.cfg
+++ b/config/hardware/probes/bltouch.cfg
@@ -14,131 +14,15 @@ gcode:
 
 [bltouch]
 sensor_pin: ^PROBE_INPUT
-#   Pin connected to the BLTouch sensor pin. Most BLTouch devices require a
-#   pullup on the sensor pin (prefix the pin name with "^").
-#
-#   This parameter must be provided.
-
 control_pin: SERVO_PIN
-#   Pin connected to the BLTouch control pin.
-#
-#   This parameter must be provided.
-
-# pin_move_time: 0.680
-#   The amount of time (in seconds) to wait for the BLTouch pin to move up or
-#   down.
-#
-#   The default is 0.680 seconds.
-
-# stow_on_each_sample: True
-#   This determines if Klipper should command the pin to move up between each
-#   probe attempt when performing a multiple probe sequence.
-#
-#   Read the directions in docs/BLTouch.md before setting this to False.
-#
-#   The default is True.
-
-# probe_with_touch_mode: False
-#   If this is set to True then Klipper will probe with the device in
-#   "touch_mode".
-#
-#   The default is False (probing in "pin_down" mode).
-
-# pin_up_reports_not_triggered: True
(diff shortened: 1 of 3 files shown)


Surion79:
I think you can cut the comments out like in all the other probe files. (You can't edit the file there anyway)
Otherwise it looks fine for me

agios:
Functionality wise looks good!

nmsmith22389:
> I think you can cut the comments out like in all the other probe files. (You can't edit the file there anyway)
> 
> Otherwise it looks fine for me

I figured... _(I have a tendancy to over annotate 🤓)_

I'll get rid of them and push the changes.

Frix-x:
I'll wait for your changes and then I'll merge! So I assume that at the end, no changes were needed in the generic_probe.cfg macro for the specificity of the BLTouch?

Thanks for your PR :)

tristanmcpherson:
Maybe add a comment outlining that if they need to define a safe homing position, it will by default use zero_reference_position if defined on the bed mesh, otherwise it will try to z home at the center. The reason I bring this up is a bit tangential, but when I was setting up BLTouch I crashed my hotend a few times because it was homing at 0, 0 due to my improper setting of zero_reference_position.

tristanmcpherson:
Do we also want to add support for `ACTIVATE_PROBE` and `DEACTIVATE_PROBE`?

Frix-x:
> Do we also want to add support for `ACTIVATE_PROBE` and `DEACTIVATE_PROBE`?

Yes, if the BLTouch needs some specific actions to be done before it can probe, if this is not the case, there is nothing to add in here.

In fact, these macros are called everywhere in Klippain whenever probing is needed. In the case of a dockable probe like a klicky, it will call the take probe and dock probe macros. In the case of a TAP that needs to probe at a safe temperature, it will lower the temperature as needed and raise it again afterwards. For an inductive probe, nothing specific is needed, so the macro doesn't have anything that belongs to that type of probe.

Since I'm not very aware of how a BLTTouch work exactly, I'm not sure if anything is needed in these macros. Maybe to deploy the needle? Not sure if this is done automatically by Klipper or not.

agios:
The needle is indeed automatically deployed by Klipper, no further actions should be needed in activate/deactivate. Both z endstop and tilt have been working fine for a week for me without any modification.

nmsmith22389:
Yeah, Klipper handles all of that so it should all be fine as is.

I use these changes on my printer with a BLTouch and everything works as intended.

The only thing I'm not entirely sure about is Klippain compatibility/interactions with `[safe_z_home]`. Specifically options like `home_xy_position` and `z_hop`. I haven't messed around with it yet.

Frix-x:
> The only thing I'm not entirely sure about is Klippain compatibility/interactions with `[safe_z_home]`. Specifically options like `home_xy_position` and `z_hop`. I haven't messed around with it yet.

This is ok as it is handled by the homing_override Klipper section and I think the behavior should be pretty close to that of the inductive probe :)

Frix-x (approved):

PR #452, opened 2024-02-04
next version

Frix-x:
# Klippain v4.3.0

As always, to get the latest features, look at the new printer.cfg, variables.cfg, etc... from the [user_template](https://github.com/Frix-x/klippain/tree/main/user_templates) folder and backport the necessary changes in your own files. There should be no real breaking changes on this aspect in this update, so this should not be mandatory.

## Special Note: Shake&Tune Module and HappyHare v2 integration :rocket:

First highlight: Shake&Tune is now properly integrated into Klippain as a real and independent module. This means that you can now update one or the other independently. But to allow this, there is a small breaking change: you have to restart the install script again to migrate your current install:
```bash
wget -O - https://raw.githubusercontent.com/Frix-x/klippain/main/install.sh | bash
```

But that's not the only big work of this release: there is now also a premium support of HappyHare v2 in Klippain to allow the use of any MMU easily and natively in Klippain!

### Configuration and Templates:
- Added Mellow SB2040 v2 MCU template
- Added Mellow SB2040 Pro MCU template
- Added template for Fysetc S6 V2.x MCU
- Added Mellow super8 MCU template
- Added Fysetc SB Can TH board template
- Added BTT Kraken MCU template
- Added BTT MMB CAN MCU template for MMUs
- Added LDO Leviathan MCU template
- Added BTT SKR2 MCU template
- Added LSO Nitehawk MCU template
- Added Octopus Pro v1.1 MCU template
- Added Octopus Pro v1.0 MCU template (as a dummy template that reference the original Octopus MCU template)
- Added Fysetc Catalyst MCU template
- Added TR8x2 leadscrew config template for Trident printers
- Added a config file for older Fysetc mini12864 v1.2 and v2.0 displays
- Added support for the V0 display MCU
- Added full support of the classic display with the creation of a proper menu with special entries for Klippain macros and a very cool boot logo!
- Finally added BLTouch support to Klippain for a proper management of the Ender3 and other popular printers
- Added support for custom thermistors type in Klippain. The first one officially added is CMFB103F3950FANT used on some LDO boards
- Fixed SPI bus pins on the BTT Manta M8P template to properly allow using SPI TMC drivers such as the TMC2240
- Fixed the M3 driver step pin on the BTT Manta M8P v2 template alongside the fan pins and properly label the CAN H/L ports
- Added support for TMC5160 drivers on the Z axis. Update your `mcu.cfg` to get the additional lines if needed
- Fixed fan pins on Fysetc Spider v3.x MCU template
- Fixed the rsense value that was wrong for the TMC5160 includes. This is probably a breaking change if you were using it as it will change the resulting motor currents!
- Added Sherpa mini extruder config file
- Added support for BTT S2DW v1.0 and similar RP2040 USB accelerometer boards. Be careful as LIS2DW is known to be problematic for proper input shaper calibration, but this config file allows you to use it even though it is not recommended
- Added support for part fan and hotend fan tachometers with their own automatic monitoring macros during printing.
- Big rework was done on the led management to allow official support for the led effects plugin
- Added rainbow barf support and new fancy led configurations

### General macros and scripts changes:
- Filter is no longer enabled to recycle and filter the chamber atmosphere at the end of a print if it was not used during the print. 
- `PARK` macro now allows X and Y parameters in case you want to use it manually or in one of your macros to park the toolhead in a specific position.
- Filament sensor usage is now managed per material configuration from your `variables.cfg`. If you don't specify it there, it will be enabled by default if available, but you should update your `variables.cfg` to the latest version to avoid problems.
- The `TIP_SHAPING` call has been removed from the `END_PRINT` and `CANCEL_PRINT` macros to eliminate the small drips that can occur at the end of a print, which can be problematic for some of you.
- The filament sensor management has been improved to be simpler and more universal, and to correctly manage the deactivation/reactivation when needed (during filament loading and unloading, etc...). This change should be transparent to you.
- Avoid saying "no heatsoak needed" when this action is disabled by the user to avoid confusion.
- Fixed a bug with the `STATUS_LEDS` on printer startup.
- Prime line has been improved to be adaptive and now follows the print on the bed. This allows for better adhesion since it is closer to the mesh. Some new parameters are also available in `variables.cfg` if you want to customize it a bit more.
- Fixed a small bug that prevented a retract in the END and CANCEL print macro in some case with an MMU configured and if the nozzle temperature was too low
- Added some logs on idle timeout to make it more clear that this timeout was reached
- Added support for dockable probe with a separate retainer such as the KlackEnder probe
- Better management of the nozzle temperature to avoid cooling it if it was already hot before starting a print when not running a probe type that need it
- Added a `PARK_FRONT` macro for easy toolhead maintenance

### System and Installation:
- The Shake&Tune module is now properly installed as a standalone module in Klippain to allow for independent update management.
- A major rework of the MMU integration has been done to allow native use of HappyHare v2.
- Klippain is now compatible with Klipper's `SET_PRINT_STATS_INFO`. To use it, add `TOTAL_LAYER=[total_layer_count]` to your slicer custom start gcode to tell Klipper the total number of layers. According to the official documentation, you also need to add the corresponding `SET_PRINT_STATS_INFO CURRENT_LAYER={layer_num}` to your slicer custom layer change gcode.
- Added an automatic GitHub action to automatically filter, tag, and eventually close outdated issues and pull requests.
- Changed the definition of Moonraker's update manager section for Klippain to remove the "beta" channel and allow it to fetch all commits instead of just the tagged releases.

### Miscellaneous:
- Added a default TPU material section to the `variables.cfg` file.
- Fixed a typo in the spelling of the word "override" in some Klippain config file names and folders.


Surion79:
Don't forget the stale action pr :)

PR #454, opened 2024-02-05
Fixed broken macros for nozzle and status led on/off. 

claudioguareschi:
Still need to fix SET_LED macro to disable effects if leds are set when effects are running


install.sh
Frix-x, line 1: Can you reset this file to the original for the PR?

moonraker/base.conf
Frix-x, line 1: Same here: can you reset this file to the original for the PR?

Frix-x (changes requested):
Thanks for the fix and for following up on this issue. Your contributions are greatly appreciated :)

I just added a small change request to be able to merge the PR cleanly as everything else looks good to me.

install.sh
claudioguareschi: Done

moonraker/base.conf
claudioguareschi: done

install.sh
```diff
@@ -27,7 +27,7 @@ BACKUP_PATH="${HOME}/klippain_config_backups"
 # Where the Klipper folder is located (ie. the internal Klipper firmware machinery)
 KLIPPER_PATH="${HOME}/klipper"
 # Branch from Frix-x/klippain repo to use during install (default: main)
-FRIX_BRANCH="main"
+FRIX_BRANCH="develop"
```
Frix-x, line 30: Sorry, you just miss this last line and it will be good to go ^^

claudioguareschi: done

PR #455, opened 2024-02-06
Allow to install Klippain in a given directory

arcadien:
Allow to install Klippain for a given Kiauh installed printer.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,9 +3,11 @@
 ###### AUTOMATED INSTALL AND UPDATE SCRIPT ######
 #################################################
 # Written by yomgui1 & Frix_x
-# @version: 1.3
+# @version: 1.5
 
 # CHANGELOG:
+#   v1.5: - Allow multi printer. Path of 'printer_data' can be passed as a script parameter. If parameter is not present,
+#           `${HOME}/printer_data/` is used as default value.
 #   v1.4: added Shake&Tune install call
 #   v1.3: - added a warning on first install to be sure the user wants to install klippain and fixed a bug
 #           where some artefacts of the old user config where still present after the install (harmless bug but not clean)
@@ -18,8 +20,6 @@
 #   v1.0: first version of the script to allow a peaceful install and update ;)
 
 
-# Where the user Klipper config is located (ie. the one used by Klipper to work)
-USER_CONFIG_PATH="${HOME}/printer_data/config"
 # Where to clone Frix-x repository config files (read-only and keep untouched)
 FRIX_CONFIG_PATH="${HOME}/klippain_config"
 # Path used to store backups when updating (backups are automatically dated when saved inside)
@@ -29,6 +29,17 @@ KLIPPER_PATH="${HOME}/klipper"
 # Branch from Frix-x/klippain repo to use during install (default: main)
 FRIX_BRANCH="main"
 
+FORCE_PRINTER_NAME=$1
+
+# Default configuration top directory
+TARGET_DIR=${HOME}/printer_data
+
+# Where the user Klipper config is located (ie. the one used by Klipper to work)
+USER_CONFIG_PATH="${TARGET_DIR}/config"
+
+# Default values
+PRINTER_NAME=printer
+SERVICE_NAME=klipper.service
 
 set -eu
 export LC_ALL=C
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -241,18 +268,18 @@ function install_mcu_templates {
...
 
-BACKUP_DIR="${BACKUP_PATH}/$(date +'%Y_%m_%d-%H%M%S')"
 
 printf "\n======================================\n"
 echo "- Klippain install and update script -"
 printf "======================================\n\n"
 
 # Run steps
 preflight_checks
+BACKUP_DIR="${BACKUP_PATH}/$(date +'%Y_%m_%d-%H%M%S')"
```
Frix-x, line 282: Is it possible to add the printer name in the backup name?

Frix-x (commented):
Thanks for this PR!

This is something that was indeed needed and that I planned to work on at some point so thank you very much!

My only concern is about the cloned klippain repository: if you update one printer, all the Klippain printers of the host will be updated as well. But I think it's a good starting point already and not necessarly a problem if it's stated in the documentation.

As a side note, the [Shake&Tune](https://github.com/Frix-x/klippain-shaketune) package will also need to be updated to be able to follow this change. Do you want to do a PR here as well?

install.sh
arcadien: Hello, I tried to pack all changes at the same place. `BACKUP_DIR` is updated (printer name is append) at line 59.
I may propagate to [Shake&Tune](https://github.com/Frix-x/klippain-shaketune) once I do it for my CR-10.

arcadien: I discover an issue with the patch. `printer_data` is also hardcoded in `shell_commands.cfg`.  I had to manually update.
The file is a symlink from Klippain installation directory, and hence is shared across all printer configuration.

arcadien: Ok done

tsk-2222:
I did a kiauh multi install of 10 printers early last year. At the time, the installed directories were named printer_1_data to printer_10_data.  It also created klipper-1.service to klipper-10.service.

When I try to run arcadien's install script with this command: './install.sh 8', the script fails because the '8_data' directory does not exist.

When I try to run the install script with this command: './install.sh printer_8', the script fails because it is looking for klipper-printer_8.service which doesn't exist.

Are you guys seeing this or do I have a 'funny' installation?

tsk-2222:
@arcadien I can't test the kiauh scripts now but it looks like my situation happens when you tell kiauh how many printer instances to setup and DO NOT assign custom names.  In this case,  **# by KIAUH convention, all instance names of only numbers need to be prefixed with 'printer_'**

You can see this in the kiauh utilities.sh at line 700.
`if [[ ${name} =~ ^[0-9]+$ ]]; then
        cfg_dirs+=("${HOME}/printer_${name}_data/config")
      else
        cfg_dirs+=("${HOME}/${name}_data/config")`

arcadien:
Ok so in that case, there will be `printer_xx_data` folder, but `klipper-xx.service`

arcadien:
@tsk-2222 it should be fixed by last commit

Frix-x:
What is the state of this PR. Is it ok now? @tsk-2222 can you confirm it's ok for you now?
Unfortunately, I can only test it on a "single printer" host on my side.

tsk-2222:
It's still broken on my end.  Something like the following snippet seems to work for the Klippain install. It just strips off any leading "printer_" characters from SERVICE_NAME and allow the script to be called as such: "./install.sh printer_8"

```
if [ "$FORCE_PRINTER_NAME" != "" ]; then
      if [ -d "${HOME}/${FORCE_PRINTER_NAME}_data" ]; then
        echo "[PRE-CHECK] Installing Klippain for printer '${FORCE_PRINTER_NAME}'"
        PRINTER_NAME=$FORCE_PRINTER_NAME
#        if [[ ${PRINTER_NAME} =~ ^[0-9]+$ ]]; then
#          TARGET_DIR=${HOME}/printer_${PRINTER_NAME}_data
#        else
          TARGET_DIR=${HOME}/${PRINTER_NAME}_data
#        fi
        USER_CONFIG_PATH="${TARGET_DIR}/config"
        SERVICE_NAME=klipper-${PRINTER_NAME#printer_}
```
The call to the shaketune install script will of course still fail though. I'm not sure how one would include parameters on the wget line for the shaketune install script.

tsk-2222:
I made some changes to install.sh and I believe it should work but I would some more eyeballs.

https://github.com/tsk-2222/klippain/commit/dfd8e6100666c538ca3d615bdcc517d38b1a5f50

stalsma:
Not sure if this project is being maintained anymore, but I stumbled upon this issue, and hacked a fix together.  Just leaving a note in case anyone else runs into the same issue.

The problem is that KIAUH takes a printer name and prefixes it with `printer_` for the data files, and but `klipper-` for the service.  The script as written will always break when a name is passed in, unless that name starts w digit, b/c it follows different prefixing conventions.  Specifically, it expects the printer name to contain `printer_` as a part of the name, but fails to remove it when checking for the klipper service on that name (b/c the service isn't prefixed with 'printer_').

So to get it to work, I simply changed the assumption, and pass in a name (e.g. myName) and then prepend the printer/klipper prefixes to that string when assigning `TARGET_DIR`.

Hope this helps someone.

```
# Step 1: Verify that the script is not run as root and Klipper is installed.
#         Then if it's a first install, warn and ask the user if he is sure to proceed
function preflight_checks {
    if [ "$EUID" -eq 0 ]; then
        echo "[PRE-CHECK] This script must not be run as root!"
        exit -1
    fi

    if [ "$FORCE_PRINTER_NAME" != "" ]; then
      if [ -d "${HOME}/printer_${FORCE_PRINTER_NAME}_data" ]; then
        echo "[PRE-CHECK] Installing Klippain for printer '${FORCE_PRINTER_NAME}'"
        PRINTER_NAME=$FORCE_PRINTER_NAME
        if [[ ${PRINTER_NAME} =~ ^[0-9]+$ ]]; then
          TARGET_DIR=${HOME}/printer_${PRINTER_NAME}_data
        else
          TARGET_DIR=${HOME}/printer_${PRINTER_NAME}_data
        fi
        USER_CONFIG_PATH="${TARGET_DIR}/config"
        SERVICE_NAME=klipper-${PRINTER_NAME}
        BACKUP_PATH=${BACKUP_PATH}/${PRINTER_NAME}
      else
        echo "[PRE-CHECK] target directory '${HOME}/${FORCE_PRINTER_NAME}_data' does not exist."
        exit -1
      fi
    fi
```

mjonuschat:
I'm going to close this and the sibling PRs with similar functionality out. There is no good standard for multi-instance klipper setups and supporting them would require a full re-architecture of the installer and possible Klippain subsystems, not just a "piecemeal" approach.

PR #458, opened 2024-02-07
feat: Provide mapping and template hardware for Creality printers

arcadien:
So far, there will be stock components for the CR-10, and the v2.1 board definition.


henrivdr (changes requested):
Should action_raise_error be inside the verbose flag?
We would want the validations regardless of verbosity.

mjonuschat:
Closing this out as it's a 2+ year old draft PR

PR #462, opened 2024-02-09
added fix to prevent issues when updating Klippain

Surion79:
When trying the new develop branch I found out that there is an issue for people not using MMU, not passing MMU parameters and not updating variables.cfg manually. (Basically just updating Klippain through Moonraker)

My take on this is to check for MMU integration and then, if not, prevent using "default" for non-existing variables.

This makes klippain usable after updates if no new features (here MMU) is used.

This is a fix, until some sort of updater for variables.cfg is implemented.


Surion79:
I also removed a variable which is not used in start_print (was moved to the _MMU_INIT functions in a different file)

Surion79:
> Are you sure it's work like that to pass param from slicer?
> It seemed to me that I had tried and that it was a problem.
> 
> What if we put the variables in the hardware mmu.cfg file and the variables.cfg file is possibly used to override the variables?

I just check if klippain is used and then set default params values. The only option otherwise possible would be to exclude check gates variable set to not set a check gates default param. But it doesn't matter this way since it is never used.

I'm against breaking up variables.cfg for this issue since the fix is only for 'old' installations, which are updating existing installations and not using ercf/mmu.
As described initially, if we implement a variables.cfg upgrader (add missing variables) the fix can be reverted.

But at the current stage it is a must have or it will break 99% of Klippain installations when pushed to main without this fix.

Surion79:
resolved conflicts

macros/base/start_print.cfg
```diff
@@ -25,7 +27,11 @@ gcode:
     {% set INITIAL_TOOL = params.INITIAL_TOOL|default(0)|int %} # Initial tool (for the MMU/ERCF initialization)
     {% set MATERIAL = params.MATERIAL|default(printer["gcode_macro _USER_VARIABLES"].print_default_material)|string %} # Material type set in the slicer
     {% set FL_SIZE = params.SIZE|default("0_0_0_0")|string %} # Get bounding box of the first layer for the adaptive bed mesh
-    {% set CHECK_GATES = params.CHECK_GATES|default(printer['gcode_macro _USER_VARIABLES'].mmu_check_gates_on_start_print)|int %} # Check if MMU gates are availables
+    {% if klippain_mmu_enabled %}
+        {% set CHECK_GATES = params.CHECK_GATES|default(printer['gcode_macro _USER_VARIABLES'].mmu_check_gates_on_start_print)|int %} # Check if MMU gates are availables
+    {% else %}
+        {% set CHECK_GATES = 0 %}
```
Benoitone, line 33: Not sure if this is necessary :)
and it's no use...

Surion79: It is, because of the following

`SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=check_gates VALUE={CHECK_GATES}`

Surion79: i only added it because if i don't set CHECK_GATES the SET_GCODE_VARIABLE will fail.
I intially had it in until the error came up. so it is a fix of a followup error.

macros/base/end_print.cfg
```diff
@@ -6,7 +6,6 @@ gcode:
     {% set safe_extruder_temp = printer["gcode_macro _USER_VARIABLES"].safe_extruder_temp|float %}
     {% set light_intensity_end_print = printer["gcode_macro _USER_VARIABLES"].light_intensity_end_print %}
     {% set klippain_mmu_enabled = printer["gcode_macro _USER_VARIABLES"].klippain_mmu_enabled %}
```
Benoitone, line 8: did you try like that?
```suggestion
    {% set klippain_mmu_enabled = printer["gcode_macro _USER_VARIABLES"].klippain_mmu_enabled %}
    {% set CHECK_GATES = params.CHECK_GATES|default(printer['gcode_macro _USER_VARIABLES'].mmu_check_gates_on_start_print,0) %} # Check if MMU gates are availables
```

macros/base/end_print.cfg
```diff
@@ -17,6 +16,7 @@ gcode:
     PARK
 
     {% if klippain_mmu_enabled %}
+        {% set mmu_unload = params.MMU_UNLOAD_AT_END|default(printer["gcode_macro _USER_VARIABLES"].mmu_unload_on_end_print)|int %}
```
Benoitone, line 19: and remove this one

macros/base/end_print.cfg
Surion79: that doesn't make sense. you added source code that i removed because it throws an error. (line 8)
regarding line 9: why haven't you added it in the first place like this? ;)

macros/base/end_print.cfg
Surion79: why remove it? does that mean we don't use MMU_UNLOAD_AT_END parameter at all? Then we can remove all references of it in Klippain

macros/base/end_print.cfg
Surion79: default on non existing variable errors out with dict error unknown. so that idea is not working

Surion79:
@Benoitone your ideas didn't work. either it is a jinja or klipper restriction

macros/base/end_print.cfg
Benoitone: I make a mistake between start_print and end_print...
```suggestion
    {% set mmu_unload = params.MMU_UNLOAD_AT_END|default(printer["gcode_macro _USER_VARIABLES"].mmu_unload_on_end_print,0) %}
```

Benoitone:
```suggestion
    {% set klippain_mmu_enabled = printer["gcode_macro _USER_VARIABLES"].klippain_mmu_enabled %}
    {% set mmu_unload = params.MMU_UNLOAD_AT_END|default(printer["gcode_macro _USER_VARIABLES"].mmu_unload_on_end_print,0) %}
```

Surion79:
@Benoitone I did not add the cancel print default, since there is no default there which requires additional defaults.

PR #463, opened 2024-02-09
Mellow fly super8

Benoitone:
add Mellow Fly Super8 mcu board


RichardTHF (approved):
Matches board pins for  Super8

Frix-x (approved):

PR #466, opened 2024-02-10
Make prime line height a variable parameter

arcadien:
Needed to allow PRIMELINE to work on my setup.


Surion79:
Looks good.
Since we do not have an updater for the variables.cfg, can you please make sure that if someone has not the variable defined after update, that the user can still expect the default value of 0.6 been used? Thanks :)

Surion79:
It misses way more functional safety issue. like preventing settings like 0.0, -4.0, 100.0

arcadien:
> Since we do not have an updater for the variables.cfg, can you please make sure that if someone has not the variable defined after update, that the user can still expect the fixed default value of 0.6 been used? Thanks :)

Done, have to test it with Klipper fully up

Surion79 (approved):
looks good for me

Surion79 (approved):
looks good for me on develop ;)

Frix-x (approved):

PR #467, opened 2024-02-10
Update base.conf

smwoodward:
Changed Primary Branch


Surion79:
Please use develop branch, since it has now all features of the HHv2 branch. The HHv2 branch is outdated

PR #468, opened 2024-02-10
removed z_offsets from read-only files

smwoodward:
updated z_offsets as discussed here https://github.com/Frix-x/klippain/pull/396


Surion79:
why now a second PR? You correctly linked one related to the change. the discussion is still not resolved since your change might make klippain unstartable because of missing configuration.

smwoodward:
I just closed it.  The other PR wasn't correct for what we were talking about doing.  

Would suggest uncommenting the offsets by default in the overrides.cfg or leave them commented out as they currently are?

PR #470, opened 2024-02-11
Fix Manta M8P V2.0 Motor SPI port

reineruhry:
Fixed the Manta M8P V2.0 Motor Port


reineruhry:
Canceled PR #389 fixes it.

PR #471, opened 2024-02-11
Add Feature absolute extrusion mode selection

reineruhry:
Added a variable defaulted to False, to allow selection between absolute and relative extrusion.

Default value: 
variable_extrusion_absolute_mode: False

To enable absolute extrusion mode set it to:
variable_extrusion_absolute_mode: True

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -46,6 +46,7 @@ gcode:
     {% set bed_mesh_enabled = printer["gcode_macro _USER_VARIABLES"].bed_mesh_enabled %}
     {% set firmware_retraction_enabled = printer["gcode_macro _USER_VARIABLES"].firmware_retraction_enabled %}
     {% set filter_enabled = printer["gcode_macro _USER_VARIABLES"].filter_enabled %}
+    {% set extrusion_absolute_mode = (printer["gcode_macro _USER_VARIABLES"].extrusion_absolute_mode|default(false)) %}
 
     {% if MATERIAL not in printer["gcode_macro _USER_VARIABLES"].material_parameters %}
         RESPOND MSG="Material '{MATERIAL}' is unknown!"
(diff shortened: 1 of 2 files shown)


arcadien:
Good idea. I noticed usage of PRIMELINE forces M83. An option is a good deal for one who wants to stay with M82

reineruhry:
> Good idea. I noticed usage of PRIMELINE forces M83. An option is a good deal for one who wants to stay with M82

I kept everything else as it was, just moved back to M82 at the end, keep PRIMELINE calculation as it's.

Surion79:
so, when slicer is configured for relative (G92 0 on layer change) and force it via klippain to absolute. what will be the result?
The slicer shows errors if you do incompatible stuff.

reineruhry:
> so, when slicer is configured for relative (G92 0 on layer change) and force it via klippain to absolute. what will be the result? The slicer shows errors if you do incompatible stuff. For me this is a slicer setting which should stay there.

If you're slicer is not configured for Relative printing, it'll fail complaining about too much extrusion, my case. Took a while to find that the extrusion mode was Relative, and I use absolute.

Surion79:
> > so, when slicer is configured for relative (G92 0 on layer change) and force it via klippain to absolute. what will be the result? The slicer shows errors if you do incompatible stuff. For me this is a slicer setting which should stay there.
> 
> If you're slicer is not configured for Relative printing, it'll fail complaining about too much extrusion, my case. Took a while to find that the extrusion mode was Relative, and I use absolute.

and not adding a M83/M82 in your default start code in your slicer config would suffice? I am not sure about this change, since there is a slicer setting for it.

Surion79:
https://www.sublimelayers.com/2017/10/to-extruder-relative-or-not-to-extrude.html#:~:text=In%20general%2C%20it%20is%20best%20practice%20to%20use,the%20code%20for%20relative%20extrusion%20mode%20is%20M83.

image

reineruhry:
Hi,

I was using this configuration to keep my already sliced compatible, avoiding to re-slice my whole "library".

May be also is valid to add a note to the documentation, mentioning that the Klippain is by default Relative Extrusion.


Best regards,
Reiner

Surion79:
For you i could think of overriding a custom start action and add your absolute extrusion there.
I think it is not wise to have klippain set something different as the slicer software

reineruhry:
Re-slicing everything!

PR #480, opened 2024-02-17
fix for extrude when temperature is too low

Benoitone:
just to avoid error when extruder temperature is too low in rare case on cancel and end print macros


Surion79:
Also on error in start_print 😉

Benoitone:
> Also on error in start_print 😉

where??

Surion79:
> > Also on error in start_print 😉
> 
> where??

the other thing with missing mmu variables in defaults. it provoked same error message regarding "can't extrude"

Frix-x (approved):

PR #481, opened 2024-02-17
For FixMmuErrorOnPrintStart branch

Benoitone:
With this it seems to work on my side.
@Surion79 Can you have a look?


Surion79:
I tried that and it didn't work. Defaulting to a non-existant variable throws a dict error. Adding a ',0' behind it won't work because macros are parsed in jinja completly prior to executing them. So while preparing the macro it errors out.

PR #482, opened 2024-02-18
Fix mmu error on print start

Benoitone:
@Surion79 can you have a look at this fix...
I think it's easier than yours?
I have test with and without the mmu variables in variables.cfg


Surion79:
the parts required, I added manually. Thanks for the alterantive :)

PR #483, opened 2024-02-19
Support Rainbow Barf style LEDs with the current LED macro implementation

gralves2:
I know Rainbow Barf toolhead support is scheduled for next version (#164) but I figured out a simple way to add support for it while keeping the current macro implementation. It has no support for animations but it works as expected (logo and nozzle leds changes color according to Status). All I did was change the chain count, add the correct RGB ordering and changed the index variables with the correct indexes for each led.


Surion79:
i am not sure, but the current develop branch will be the next release in a couple days. so i am not sure, how this PR would get ahead of this schedule? Is it required with the new version?

gralves2:
I just checked the development branch and I think this pull request is redundant with it.

Issue #493, opened 2024-02-23
Enable goto purge bucket and maintenance in GUI


elpopo-eng:
### Describe the feature or hardware support you'd like

Enable goto purge and maintenance in GUI

PR #494, opened 2024-02-23
Add park front

elpopo-eng:
Enable goto purge bucket and maintenance in GUI #493

### DIFF
--- a/macros/base/park.cfg
+++ b/macros/base/park.cfg
@@ -53,8 +53,8 @@ gcode:
     {% endif %}
 
 
-[gcode_macro GOTO_MAINTENANCE]
-description: Move over the purge bucket
+[gcode_macro PARK_FRONT]
+description: Park the toolhead on the front of the printer for maintenance
 gcode:
     {% set St = printer["gcode_macro _USER_VARIABLES"].travel_speed * 60 %}
     {% set Sz = printer["gcode_macro _USER_VARIABLES"].z_drop_speed * 60 %}
@@ -62,8 +62,8 @@ gcode:
     {% if "xyz" not in printer.toolhead.homed_axes %}
         G28   ; home if not already homed
     {% endif %}
-    SAVE_GCODE_STATE NAME=GOTO_MAINTENANCE
+    SAVE_GCODE_STATE NAME=PARK_FRONT
     G90      ; absolute positioning
     G0 Z{printer.toolhead.axis_maximum.z/3*2} F{Sz}
     G0 X{printer.toolhead.axis_maximum.x/2} Y{printer.toolhead.axis_minimum.y+10} F{St}
-    RESTORE_GCODE_STATE NAME=GOTO_MAINTENANCE
+    RESTORE_GCODE_STATE NAME=PARK_FRONT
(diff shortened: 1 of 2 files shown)


Frix-x:
Here is my remarks as I'm not a very fan of this PR:

1. Maybe the maintenance could be ok, but I would rename it to "PARK_FRONT" or something like that.
1. The second macro about the purge bucket is not a good idea IMHO since not everyone is equipped with this type of bucket and the macro is included all the time, even for them. It would not be harmless since it would do "nothing" but it will still be there and cause questions. I would rather you keep it in your overrides. Also FYI, my park position is above the purge bucket and this way when I call the standard PARK macro, it effectively goes to the purge bucket. Maybe you can do the same?

elpopo-eng:
OK, I understand your position on the bucket, I'll delete it.
I've also renamed the macro to PARK_FRONT

Surion79:
i strongly recommend putting that PARK_FRONT macro in a dedicated file under macros folder and add a include in the printer.cfg, which can be uncommented, if required.

elpopo-eng:
I'm not agree with you @Surion79 
It's not a breaking change update and it's just another park method. So for me it should remain in the park.cfg file as for heatsoak.cfg, filament_swap.cfg etc...

Surion79:
it is not a breaking change. it is just adding a specific macro visible for every installation. As @Frix-x described it already.
my proposal fixes that quite easily.

macros/base/park.cfg
```diff
@@ -51,3 +51,19 @@ gcode:
...
+
+
+[gcode_macro PARK_FRONT]
+description: Park the toolhead on the front of the printer for maintenance
+gcode:
+    {% set St = printer["gcode_macro _USER_VARIABLES"].travel_speed * 60 %}
+    {% set Sz = printer["gcode_macro _USER_VARIABLES"].z_drop_speed * 60 %}
+
+    {% if "xyz" not in printer.toolhead.homed_axes %}
+        G28   ; home if not already homed
```
Surion79, line 63: I would rather do it same as original PARK behaviour and not have two different PARKs behaving differently on homed axis.

Surion79 (changes requested):
Should be ok if axis home is implemented same as in park

macros/base/park.cfg
elpopo-eng: OK for adding a house to `PARK`,
`PARK` is not really well coded, because if you add the `E` parameter when you call `PARK`, it is replaced by the default value from the restract hardware or firmware configuration, which doesn't make sense. If you force a parameter in the call, you have to apply it.
This should be part of another PR that would make `PARK` cleaner

Surion79: well, this changes it. 
Park was: Not homed -> error
Park_front was: Not homed -> homing

instead of 
park_front: Not homed -> error

it is not 
park: not homed -> homing.

I am not sure which way is the better one.
Guess something for @Frix-x to decide.

Frix-x (approved):

macros/base/park.cfg
Frix-x: It's ok for me as is. We will work on improving it in another PR if needed

PR #497, opened 2024-02-23
Update printer.cfg

Benoitone:
I need to change the order for included lines for mmu to be able to override later some mmu macros. And add future support for optional EREC Filament Cutter according to variables (actualy develop) branch of HH


Surion79 (approved):
this only affects new installations

PR #498, opened 2024-02-24
Update primeline so that it runs along the rectangle of the first layer

elpopo-eng:
Thanks to @fbeauKmi 

Now `prime_line_xy` defines the origin of the baseline
The primeline is made on the corner of the `FL_SIZE` rectangle closest to the origin.

Add Variable `prime_line_margin` wich define distance of purge line from fl_size rectangle

If `FL_SIZE` is not defined, old primeline is used

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -29,7 +29,7 @@ gcode:
     {% set TOOLS_USED = params.TOOLS_USED|default("")|string %} # Check if MMU gates (used in gcode file) are availables
     {% set SYNC_MMU_EXTRUDER = params.SYNC_MMU_EXTRUDER|default(0)|int %} # set MMU gear motor and extruder synchronization during print TODO
     {% set BED_MESH_PROFILE = params.MESH|default("")|string %} # Bed mesh profile to load
-    {% set ADAPTIVE_PRIMELINE = params.ADAPTIVE_PRIMELINE|default(True) %} # Weither to do or not an adaptive prime line near the real print zone
+    {% set ADAPTIVE_PRIMELINE = params.ADAPTIVE_PRIMELINE|default(1)|int %} # Weither to do or not an adaptive prime line near the real print zone
 
     # Set the variables to be used in all the modules based on the slicer parameters
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_temp VALUE={BED_TEMP}
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -29,6 +29,7 @@ gcode:
     {% set TOOLS_USED = params.TOOLS_USED|default("")|string %} # Check if MMU gates (used in gcode file) are availables
     {% set SYNC_MMU_EXTRUDER = params.SYNC_MMU_EXTRUDER|default(0)|int %} # set MMU gear motor and extruder synchronization during print TODO
     {% set BED_MESH_PROFILE = params.MESH|default("")|string %} # Bed mesh profile to load
+    {% set ADAPTIVE_PRIMELINE = params.ADAPTIVE_PRIMELINE|default(True) %} # Weither to do or not an adaptive prime line near the real print zone
 
     # Set the variables to be used in all the modules based on the slicer parameters
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_temp VALUE={BED_TEMP}
@@ -43,6 +44,7 @@ gcode:
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=material VALUE='"{MATERIAL}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=fl_size VALUE='"{FL_SIZE}"'
     SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=bed_mesh_profile VALUE='"{BED_MESH_PROFILE}"'
+    SET_GCODE_VARIABLE MACRO=START_PRINT VARIABLE=adaptive_primeline VALUE={ADAPTIVE_PRIMELINE}
     
     {% if params.TOTAL_LAYER %} # total layers count (if provided by the slicer)
         SET_PRINT_STATS_INFO TOTAL_LAYER={params.TOTAL_LAYER|int}
@@ -193,7 +195,15 @@ gcode:
 
 [gcode_macro _MODULE_PRIMELINE]
 gcode:
-    PRIMELINE
+    # ----- PRIME LINE -------------------------------------------
+    {% set FL_SIZE = printer["gcode_macro START_PRINT"].fl_size %}
+    {% set ADAPTIVE_PRIMELINE = printer["gcode_macro START_PRINT"].adaptive_primeline %}
+    {% set verbose = printer["gcode_macro _USER_VARIABLES"].verbose %}
+
+    {% if verbose %}
+        RESPOND MSG="Executing a primeline..."
+    {% endif %}
+    PRIMELINE SIZE={FL_SIZE} ADAPTIVE_MODE={ADAPTIVE_PRIMELINE}
 
 
 [gcode_macro _MODULE_HEATSOAK_BED]
@@ -398,7 +408,6 @@ gcode:
             {% if verbose %}
                 RESPOND MSG="Bed mesh measurement..."
             {% endif %}
-
             ADAPTIVE_BED_MESH SIZE={FL_SIZE}
         {% else %}
             {% if verbose %}
--- a/macros/helpers/prime_line.cfg
+++ b/macros/helpers/prime_line.cfg
@@ -1,18 +1,46 @@
 [gcode_macro PRIMELINE]
 gcode:
-    # Macro parameters
+    # Base macro parameters
     {% set prime_line_length = params.LINE_LENGTH|default(printer["gcode_macro _USER_VARIABLES"].prime_line_length)|float %}
     {% set prime_line_purge_distance = params.PURGE_LENGTH|default(printer["gcode_macro _USER_VARIABLES"].prime_line_purge_distance)|float %}
     {% set prime_line_flowrate = params.FLOWRATE|default(printer["gcode_macro _USER_VARIABLES"].prime_line_flowrate)|float %}
-    {% if printer["gcode_macro _USER_VARIABLES"].prime_line_height %}
-      {% set prime_line_height = params.LINE_HEIGHT|default(printer["gcode_macro _USER_VARIABLES"].prime_line_height)|float %}
-    {% else %}
-      {% set prime_line_height = params.LINE_HEIGHT|default(0.6)|float %}
+    {% set prime_line_height = params.LINE_HEIGHT|default(printer["gcode_macro _USER_VARIABLES"].prime_line_height)|default(0.6)|float %}
+    {% set prime_line_adaptive = params.ADAPTIVE_MODE|default(True) %}
+    
+    # If the SIZE parameter is defined and not a dummy placeholder, we use it to do the adaptive bed mesh logic
+    {% set coordinatesFound = false %}
+    {% if params.SIZE is defined and params.SIZE != "0_0_0_0" %}
+        {% set xMinSpec, yMinSpec, xMaxSpec, yMaxSpec = params.SIZE.split('_')|map('trim')|map('int') %}
+        {% set coordinatesFound = true %}
+    {% elif printer.exclude_object is defined %}
+        {% if printer.exclude_object.objects %}
+            # Else if SIZE is not defined, we fallback to use the [exclude_object] tags
+            # This method is derived from Kyleisah KAMP repository: https://github.com/kyleisah/Klipper-Adaptive-Meshing-Purging)
+            {% set eo_points = printer.exclude_object.objects|map(attribute='polygon')|sum(start=[]) %}
+            {% set xMinSpec = eo_points|map(attribute=0)|min %}
+            {% set yMinSpec = eo_points|map(attribute=1)|min %}
+            {% set xMaxSpec = eo_points|map(attribute=0)|max %}
+            {% set yMaxSpec = eo_points|map(attribute=1)|max %}
+            {% set coordinatesFound = true %}
+        {% endif %}
     {% endif %}
 
-    # Set internal macro vars
+    # We get the default prime line position parameters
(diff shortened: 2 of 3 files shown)


Surion79:
i don't see the benefit of it except 2 seconds of travel. for me it looks overcomplicating stuff, including possible debugging.

Aside of that: This will fail after updating klippain via moonraker as it is defaulting on a non-existing variable.

elpopo-eng:
The aim is to place the primeline as close as possible to the bdemesh area.
If we place the primeline on the edge of the bed, you can engrave your PEI, but if you like that, we can introduce a variable to use the legacy baseline...

Surion79:
Engrave? The location is so near the border, it would never get printed on anyway.

Surion79:
Don't forget my other comments regarding breaking klippain on update without the default fix.

macros/helpers/prime_line.cfg
```diff
@@ -30,6 +30,19 @@ gcode:
     {% set prime_line_y = params.START_Y|default(prime_line_y)|float %}
     {% set prime_line_direction = params.LINE_DIRECTION|default(printer["gcode_macro _USER_VARIABLES"].prime_line_direction)|string|upper %}
 
+    # Redefine start position from fl_size 
+    {% set fl_size=printer['gcode_macro START_PRINT'].fl_size %}
+    {% if fl_size != "0_0_0_0" %}
+        {% set prime_line_margin = params.LINE_MARGIN|default(printer["gcode_macro _USER_VARIABLES"].prime_line_margin|float) %}
```
Surion79, line 36: This will break klippain on moonraker update if a fixed value default isn't added.
We had that topic on mmu already.

macros/helpers/prime_line.cfg
```diff
@@ -30,6 +30,19 @@ gcode:
...
+        {% set min_x, min_y, max_x, max_y = fl_size.split("_")|map('float') %}
+        {% set prime_line_x = [[prime_line_x, min_x - prime_line_margin] | max, max_x + prime_line_margin] | min %}
+        {% set prime_line_y = [[prime_line_y, min_y - prime_line_margin] | max, max_y + prime_line_margin] | min %}
```
Surion79, line 39: This verifies that x/y never goes under 0/min or over max?

Surion79 (dismissed):
The default on a non existing variable will break *all* installations which are updating existing klippain installations. Fix is to add another default with the default value. That was the solution @Benoitone suggested when we worked on the mmu verification of his branch.

fbeauKmi:
Thanks for this PR @elpopo-eng . I think this is a nice to have. IMO the purge line should always be inside the bed mesh.Then it could be possible to go lower at the end of the prime line (Z<0.05) to wipe the nozzle on the bed.  

@Surion79 If you don't intend to use this feature, you can remove prime_line from your startup sequence and then add it to your startup gcode in the slicer.

That's the beauty of Klippain being fully overridable ❤️ .

macros/helpers/prime_line.cfg
fbeauKmi: This ensure that prime line start point is inside FL_SIZE area + prime_line_margin.
The prime_line_way parameter attempt to avoid that prime_line leaves the print area in case of misconfiguration of prime line start point.

Surion79:
> Thanks for this PR @elpopo-eng . I think this is a nice to have. IMO the purge line should always be inside the bed mesh.Then it could be possible to go lower at the end of the prime line (Z<0.05) to wipe the nozzle on the bed.
> 
> @Surion79 If you don't intend to use this feature, you can remove prime_line from your startup sequence and then add it to your startup gcode in the slicer.
> 
> That's the beauty of Klippain being fully overridable ❤️ .

errr.... My Klippain tag is reading Collaborator, and i proposed a solution? I could also just push my proposal to the Klippain branch `primeline_FLSIZE_dependent`, but I won't.

Anyway, you don't override stuff by making users override existing functionality in klippain to tell them "do that in your slicer now".

fbeauKmi:
> Anyway, you don't override stuff by making users override existing functionality in klippain to tell them "do that in your slicer now".

No offense, override in Klippain if you prefer 👍

Surion79:
still a no regarding overrides, because this change contradicts existing configuration in variables.cfg. without proper parametrization or setting an enable/disable variable, the prime line location is moved away from the settings in the variables.cfg.
i just ask for a variable to disable/enable this feature. i have no idea, why this is such hassle? it is 2 lines of code and a variable, additional to the variable which is already introduced by this PR.

Frix-x:
I appreciate the discussion surrounding this PR and the thoughtful contributions from everyone involved, but please keep it friendly. To address the concerns raised, here is my point:

1. The demand for adaptive purge integrated within Klippain is something that I get everyday on Discord or in PMs. As currently there is no solution in Klippain, I push the users over using KAMP. The problem is that using KAMP in Klippain is not that easy and can easily break the adaptive mesh: this create a lot of support work for me to help them going. So I'm definitely full on to integrate this to Klippain to have something just working out of the box.

1. The initiative for developing adaptive purge has been collaboratively discussed on Discord and is supported by contributions from Elpopo and Fbeaukmi since I had no time to develop it myself. This collaboration allowed to bring this PR and is also followed by a lot of people on Discord just waiting for it to be merged.

1. The concerns about a printed line (as the purge operation is doing) outside the meshed print zone are valid, particularly for printers with warped beds like Ender3, Prusa or Switchwire. This was mitigated by using a fixed height of 0.6mm for the prime line, but it was changed in a previous commit with an adjustable height parameter, that, while enhancing flexibility, also reintroduces the risk of crashes if using a small value closer to the print layer height like 0.2mm. Positioning the purge as close as possible to the part emerges as a prudent strategy to mitigate these risks.

1. Finally, I think that having adaptive mesh enabled by default sets a precedent for incorporating adaptive purge in a similar manner. Users can still not use the option by removing the SIZE parameter. I do not see any point in doing an adaptive mesh and not an adaptive purge or vice versa. On a good bed, it will be nothing more than a shiny feature but will not have any cons, on a warped bed it can avoid a crash.

This PR represents a step forward in aligning Klippain's features with the community's evolving needs and Klipper's advancements (like the adaptive mesh that is now mainline). So I'll review and test it properly to validate that everything is working and merge it in the current state.

Surion79:
I still request to add a disable variable, so users can choose to make a bed mesh and an outside prime line.
It can be enabled by default.

@Frix-x  Since i invested a lot of time to this project, i think it is not too much to ask.

elpopo-eng:
@Frix-x we have chosen to always orient the primeline towards the centre of the bed (case 1) in order to avoid a possible end of prime_line outside the bed (case 2) and to have a simple code.
We can change this behaviour if you don't think it's correct.

case 1 : 
image

case 2 :
image

Frix-x:
Ok, I made some changes to make it a bit more flexible and cleaner:
  - There is now an `ADAPTIVE_PRIMELINE` variable that you can use from your slicer to enable/disable the "adaptive mode" of the primeline directly from the START_PRINT sequence. By default, if nothing is set, it's True, so adaptive mode will be enabled for everyone unless we force it to be disabled.
  - The `PRIMELINE' macro can now also take an `ADAPTIVE_MODE' parameter that allow you to set the behavior when calling it manually. I think this is useful because when called manually, there is no printing going on and I understand that someone would want to know where the primeline is going to be and not randomly put it where the last print was done.
  - I've extended the SIZE parameter to work as it does with the bed_mesh by also allowing the exclude objects tags to be used when they are found and when SIZE is omitted. So in this case it will default to something like KAMP.
  - I also added a very small move at the end to prevent the head from crashing back in the primeline if the slicer starts to emit a Z down move, as I noticed some actually do and this was asked multiple times by the community.

As my main printer is currently running for a couple of hours, can someone test it and let me know if everything is working ok?

Surion79:
I habe currently an open printer and don't mind testing it. Anything specific to test? Fl size near the border?

macros/helpers/prime_line.cfg
```diff
@@ -1,18 +1,46 @@
...
-      {% set prime_line_height = params.LINE_HEIGHT|default(0.6)|float %}
+    {% set prime_line_height = params.LINE_HEIGHT|default(printer["gcode_macro _USER_VARIABLES"].prime_line_height)|default(0.6)|float %}
+    {% set prime_line_adaptive = params.ADAPTIVE_MODE|default(True) %}
+    
+    # If the SIZE parameter is defined and not a dummy placeholder, we use it to do the adaptive bed mesh logic
+    {% set coordinatesFound = false %}
+    {% if params.SIZE is defined and params.SIZE != "0_0_0_0" %}
+        {% set xMinSpec, yMinSpec, xMaxSpec, yMaxSpec = params.SIZE.split('_')|map('trim')|map('int') %}
+        {% set coordinatesFound = true %}
+    {% elif printer.exclude_object is defined %}
```
fbeauKmi, line 15: IMHO, using exclude_object to define the prime line introduces risks of collision. Although there's a lot of work being done in this area, not all slicers use exclude_object for brim, skirts and purge towers. 

But if you feel that this code should be maintained :
This code is already used in Adaptive_bed_mesh, would it be interesting to take it out of these 2 macros to introduce a mechanism to populate FL_SIZE whith exclude object during START_PRINT, for example or as reverse define an exclude_object with FL_SIZE ? Having in mind the compatibility with Klippe adaptive_mesh.

macros/helpers/prime_line.cfg
```diff
@@ -1,18 +1,46 @@
...
+    # If first layer is retrieved and adaptive mode is enabled, then we replace the coordinates to do an adaptive purge
+    {% if coordinatesFound and prime_line_adaptive %}
+        {% set prime_line_margin = params.LINE_MARGIN|default(printer["gcode_macro _USER_VARIABLES"].prime_line_margin)|default(5.0)|float %}
+        {% set prime_line_x = [[prime_line_x, xMinSpec - prime_line_margin]|max, xMaxSpec + prime_line_margin]|min %}
+        {% set prime_line_y = [[prime_line_y, yMinSpec - prime_line_margin]|max, yMaxSpec + prime_line_margin]|min %}
+    {% endif %}
+
+    # Choose the way of printing the primeline (in + or -) alongside the direction to avoid going outside the bed boundaries
+    {% set axes_max = printer.toolhead.axis_maximum %}
+    {% set prime_line_way = -1 if (prime_line_direction == "X" and prime_line_x > axes_max.x/2) or (prime_line_direction == "Y" and prime_line_y > axes_max.y/2) else 1 %}
```
fbeauKmi, line 43: Can we put the params.LINE_MARGIN line with other entry params for readability ?

As discuss with @elpopo-eng in https://github.com/Frix-x/klippain/pull/498#issuecomment-1963513548 there is an edge case where primeline could be outside the mesh. 
It add a bit of complixity but I suggest this code to mitigate this situation : 
```
# We find bed center : useful to determine pime_line position and prime_line_way
    {% set center_x, center_y  = [printer.toolhead.axis_maximum.x / 2, printer.toolhead.axis_maximum.y / 2]|map("float")  %}

    # If first layer is retrieved and adaptive mode is enabled, then we replace the coordinates to do an adaptive purge
    {% if if coordinatesFound and prime_line_adaptive" %}
        {% set min_x, min_y, max_x, max_y = fl_size.split("_")|map('float') %}

        # Fix prime_line start position while FL_SIZE is not in the same bed area
        {% set prime_line_x = 2*center_x - prime_line_x 
                               if (prime_line_x > center_x and xMaxSpec < center_x) or (prime_line_x < center_x and xMinSpec > center_x) 
                               else prime_line_x %}
        {% set prime_line_y = 2*center_y - prime_line_y 
                               if (prime_line_y > center_y and yMaxSpec < center_y) or (prime_line_y < center_y and yMinSpec > center_y) 
                               else prime_line_y %}

    
        {% set prime_line_x = [[prime_line_x, xMinSpec - prime_line_margin] | max, xMaxSpec + prime_line_margin] | min %}
        {% set prime_line_y = [[prime_line_y, yMinSpec - prime_line_margin] | max, yMaxSpec + prime_line_margin] | min %}
    {% endif %}
    
    # Choose the way of printing the primeline (in + or -) alongside the direction to avoid going outside the bed boundaries
    {% set prime_line_way = -1 if (prime_line_direction == "X" and prime_line_x > center_x) or (prime_line_direction == "Y" and prime_line_y > center_y) else 1 %}
```

macros/helpers/prime_line.cfg
```diff
@@ -71,7 +74,7 @@ gcode:
     # We then compute the height to width ratio and validate that the prime line will not be too thin
     {% if (prime_line_height / line_width) >= 0.5 %} # TODO: validate this 1/2 ratio is good for all
         {% if verbose %}
-            {action_raise_error("The prime line will be too thin and will probably not stick properly to the bed. Increase its purge distance or decrease its length! Aborting...")}
+            {action_raise_error("The prime line will be too thin and will probably not stick properly to the bed. Increase its purge distance or decrease its length!")}
```
fbeauKmi, line 77: The error is raised in Verbose mode only , is it mandatory ?

macros/helpers/prime_line.cfg
Frix-x: This will be reworked later as I do plan to switch to the mainline adaptive bed mesh at some point when slicer will manage correctly the purge tower as an object. And I'll indeed work on defining fake objects definitions to be able to use it the same way in manual calibrations macros and make it transparent for an user.

macros/base/start_print.cfg
```diff
@@ -29,6 +29,7 @@ gcode:
     {% set TOOLS_USED = params.TOOLS_USED|default("")|string %} # Check if MMU gates (used in gcode file) are availables
     {% set SYNC_MMU_EXTRUDER = params.SYNC_MMU_EXTRUDER|default(0)|int %} # set MMU gear motor and extruder synchronization during print TODO
     {% set BED_MESH_PROFILE = params.MESH|default("")|string %} # Bed mesh profile to load
+    {% set ADAPTIVE_PRIMELINE = params.ADAPTIVE_PRIMELINE|default(True) %} # Weither to do or not an adaptive prime line near the real print zone
```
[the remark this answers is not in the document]
Frix-x: You are right, I'll use 0/1 since it come from the slicer and will be easier to manage this way!

macros/helpers/prime_line.cfg
Frix-x: I fixed it using your code, thanks!

Frix-x:
New push, new tests needed before a final merge :)

PR #505, opened 2024-02-26
LDO Nitehawk MCU template and tachometer management

Frix-x:


mvdveer:
> @mvdveer @akashrpatel I've reworked it a bit to have a proper integration of tachometers in Klippain (so it get externalized from the MCU template) and I also moved the temperature sensors that were already managed by standard parts ok Klippain, etc...
> 
> Can you validate that everything is still working as it should for you? I plan to merge it tonight alongside the new Klippain version

All seems ok on my end

PR #507, opened 2024-02-26
changed pin to verify using correct voltage

Surion79:
as reported in Issue, the current defaults links Tap to a 24V pin and also on the wrong pin within the provided manual.


Frix-x (approved):
Thanks for fixing it, we indeed discussed it on Discord :)
I'll merge!

PR #513, opened 2024-02-29
Fix homing in error gcode

elpopo-eng:
When an error occurs in the gcode, we call `PARK`, but following PR #494  we force a home in the `PARK` macro.
I find it strange to do home when the job has an error.
I therefore propose to do `PARK` only when the machine is already homed.

### DIFF
--- a/config/machine.cfg
+++ b/config/machine.cfg
@@ -4,7 +4,7 @@ on_error_gcode:
     {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
         STATUS_LEDS COLOR="ERROR"
     {% endif %}
-    {% if printer["gcode_macro _USER_VARIABLES"].probe_type_enabled == "dockable" %}
+    {% if printer["gcode_macro _USER_VARIABLES"].probe_type_enabled == "dockable" or printer["gcode_macro _USER_VARIABLES"].probe_type_enabled == "dockable_virtual" %}
         _PROBE_ON_ERROR_ACTION
     {% endif %}
 
--- a/config/machine.cfg
+++ b/config/machine.cfg
@@ -1,13 +1,18 @@
 [virtual_sdcard]
 path: ~/printer_data/gcodes
 on_error_gcode:
-   {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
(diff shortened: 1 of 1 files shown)


config/machine.cfg
```diff
@@ -1,13 +1,18 @@
...
-      STATUS_LEDS COLOR="ERROR"
-   {% endif %}
-   {% if printer["gcode_macro _USER_VARIABLES"].probe_type_enabled == "dockable" %}
-       _PROBE_ON_ERROR_ACTION
-   {% endif %}
-    CANCEL_PRINT
+    {% if printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
+        STATUS_LEDS COLOR="ERROR"
+    {% endif %}
+    {% if printer["gcode_macro _USER_VARIABLES"].probe_type_enabled == "dockable" %}
```
Frix-x, line 7: can you also add a reference to "dockable_virtual" as well ? I think it was forgotten and it's a good idea to add it as well.

Frix-x (changes requested):
I've got a small change to add in this PR to make it more accurate

Surion79:
@elpopo-eng thanks a lot for this PR! this has been introduced with the _CG28 in Park. good it was found early :)

config/machine.cfg
elpopo-eng: Yes ! this make sense.

beavis69:
I've have a head crash at end of print. 
I'm overriding led_effects and the new "done_printing" led effects was not in my override.
So i got an STATUS_LEDS COLOR ERROR followed by a forced homing on my printed parts...

I think this need to be commited to master ASAP

Surion79:
> I've have a head crash at end of print. I'm overriding led_effects and the new "done_printing" led effects was not in my override. So i got an STATUS_LEDS COLOR ERROR followed by a forced homing on my printed parts...
> 
> I think this need to be commited to master ASAP

it was already merged. have you not updated to the latest version?

beavis69:
I think it was merged only in develop branch but not in main, see : https://github.com/Frix-x/klippain/blob/main/config/machine.cfg

Frix-x:
I'll releasea bugfixes version today to sync develop and main

PR #515, opened 2024-02-29
Fix park with overrided param E

elpopo-eng:
The `PARK` macro did not always take into account the overloading of the macro's E parameter.

### DIFF
--- a/macros/base/park.cfg
+++ b/macros/base/park.cfg
@@ -23,6 +23,8 @@ gcode:
         {% set z_safe = max_z %}
     {% endif %}
 
+    SAVE_GCODE_STATE NAME=PARK
+
     _CG28 ; home if not already homed
 
     {% if printer.extruder.can_extrude %}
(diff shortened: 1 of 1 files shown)


Surion79 (commented):
looks good, i would recommend putting the last respond in verbose checks as it is done in all other responds

macros/base/park.cfg
```diff
@@ -27,21 +26,35 @@ gcode:
...
+                {% if verbose %}
+                    RESPOND MSG="Firmware retraction disabled, Extruder retraction = {E}"
+                {% endif %}
             {% endif %}
+        {% endif %}
+
+        {% if E is defined and E > 0 %}
             G92 E0
             G1 E-{E} F2100
+            RESPOND MSG="Extruder retraction = {E}"
```
Surion79, line 57: this one

elpopo-eng: This respond is useless, so I've deleted it. It was introduced to test my code.

Frix-x (approved):

PR #516, opened 2024-02-29
Optimise start print homing

elpopo-eng:
Add a Z homing in the _MODULE_TILTING module if force_homing_in_start_print is true, as it makes more sense to do a G28 Z just after tilting for the rest of the start print sequence.

### DIFF
--- a/macros/base/start_print.cfg
+++ b/macros/base/start_print.cfg
@@ -313,6 +313,9 @@ gcode:
     # The correct operation is automatically selected in the _TILT_CALIBRATE macro
     {% set force_homing_in_start_print = printer["gcode_macro _USER_VARIABLES"].force_homing_in_start_print %}
     _TILT_CALIBRATE FORCE={force_homing_in_start_print}
+    {% if force_homing_in_start_print %}
+        G28 Z
+    {% endif %}
 
 
 [gcode_macro _MODULE_EXTRUDER_HEATING]


Surion79 (approved):
Looks good to me. wondering why this never came up

Frix-x:
This was the case in the past, because a G28 Z is indeed mandatory after a tilting procedure to adjust the Z zero value.

However, when I switched to decoupled START_PRINT modules, I removed it to simplify things and make the modules do only what they are designed to do. There is already a Z offset module that does this and is automatically called later in all standard sequences. So this shouldn't be a problem.

If you want to do it twice (once after the tilt and once on a final Z calibration), you can override the sequence to add a second entry. Note, however, that while this is done correctly (but later), the current default sequences are optimized to avoid performing redundant operations like this multiple times.

elpopo-eng:
I understand your point of view, and yes, you shouldn't repeat the same actions over and over again. For my part, I have a slightly different startup sequence that places the toolhead in the center of the bed before the heatsoak_chamber to blow as much as possible with the part_fan and hotend_fan. so I need to have a z that's not too bad.

The start-up sequence is not flexible enough.
I was thinking of introducing a list composed of standard actions and macros that would be filled in via a macro (Init_Start_Sequence), the list could look like this: 
```
{
  "actions": [
    {
      "type": "module",
      "name": "bed_soak"
    },
    {
      "type": "module",
      "name": "extruder_preheating"
    },
    {
      "type": "module",
      "name": "chamber_soak"
    },
    {
      "type": "macro",
      "name": "Z"
    }...
  ]
}
```
Then in the start print we would just have a loop that processes the elements one by one.

EricZimmerman:
does this not allow that?

```
## The START_PRINT sequence is modular and fully customizable. A default START_PRINT sequence is auto-populated based on
## your probe choice (TAP, Dockable, Inductive), but if for some reasons you still want to modify it, please uncomment the following 3
## lines to define a new `variable_startprint_actions`. You can use any number of action or even duplicate some actions if needed.
## Available actions: "bed_soak", "extruder_preheating", "chamber_soak", "extruder_heating", "tilt_calib", "z_offset", "bedmesh", "purge", "clean", "primeline"
##
# [gcode_macro _USER_VARIABLES]
# variable_startprint_actions: "action1", "action2", ...
# gcode: # do not remove this line
```

you can do this today and even make your own

elpopo-eng:
> does this not allow that?
> 
> ```
> ## The START_PRINT sequence is modular and fully customizable. A default START_PRINT sequence is auto-populated based on
> ## your probe choice (TAP, Dockable, Inductive), but if for some reasons you still want to modify it, please uncomment the following 3
> ## lines to define a new `variable_startprint_actions`. You can use any number of action or even duplicate some actions if needed.
> ## Available actions: "bed_soak", "extruder_preheating", "chamber_soak", "extruder_heating", "tilt_calib", "z_offset", "bedmesh", "purge", "clean", "primeline"
> ##
> # [gcode_macro _USER_VARIABLES]
> # variable_startprint_actions: "action1", "action2", ...
> # gcode: # do not remove this line
> ```
> 
> you can do this today and even make your own

not really, because we can't insert simple macros, they have to be standard modules and there are only 3 custom modules.

EricZimmerman:
those are just macros getting called. you cant make more?

theres these already included:

```
  {% elif action == "custom1" %}
            _MODULE_CUSTOM1 {rawparams}
        {% elif action == "custom2" %}
            _MODULE_CUSTOM2 {rawparams}
        {% elif action == "custom3" %}
            _MODULE_CUSTOM3 {rawparams}
            ```
            
            
            and
            
            ```
            
[gcode_macro _MODULE_CUSTOM1]
gcode:
    # ---- CUSTOM Macro section
    # this section is reserved for personal customized start actions, which can be combined with all other start actions
    # in order to use this, create a new macro in overrides.cfg
    # [gcode_macro _MODULE_CUSTOM1]
    # gcode:
    #   ## Your custom code here


[gcode_macro _MODULE_CUSTOM2]
gcode:
    # ---- CUSTOM Macro section
    # this section is reserved for personal customized start actions, which can be combined with all other start actions
    # in order to use this, create a new macro in overrides.cfg
    # [gcode_macro _MODULE_CUSTOM2]
    # gcode:
    #   ## Your custom code here


[gcode_macro _MODULE_CUSTOM3]
gcode:
    # ---- CUSTOM Macro section
    # this section is reserved for personal customized start actions, which can be combined with all other start actions
    # in order to use this, create a new macro in overrides.cfg
    # [gcode_macro _MODULE_CUSTOM3]
    # gcode:
    #   ## Your custom code here


```

how many more do you need?

elpopo-eng:
these are fix modules and can't be modified without restart, with the lsite we can modify it live.
And this would be very useful for development.

EricZimmerman:
I dont think catering to everyones hyper custom requirements is feasible either.

Surion79:
> I dont think catering to everyones hyper custom requirements is feasible either. 

Exactly.
Lately some PRs gave me the feeling, that it was more about integrating hyper custom requirements instead of evolving Klippain's generic/standard way.

Introducing the custom modules was exactly the point to have the option to add custom code. In this case it is a trivial solution with using one of these modules.
I think @Frix-x decided already. Or he has to decide.

EricZimmerman:
Maybe adding two more custom modules in print start for a total of five would be a good benefit that 99.97% of users would never use but the remaining users may find useful

There's nothing stopping the .03% of users from taking one of those custom commands and turning it into 15 subcommands in their own hypercomplicated macro either

Surion79:
there was a merge to use now 9 custom modules instead of 3.

elpopo-eng:
I'm really not a fan of this ugly switch, looks like programming from 30 years ago

EricZimmerman:
Yet it works and allows for a lot of customized actions for the few that need or want it. It's not wrong if it works.

Surion79:
well, generic programming has advantages and disadvantages.

Frix-x:
> I'm really not a fan of this ugly switch, looks like programming from 30 years ago

I agree with you but Jinja is pretty limited and doesn't allow reflection programming unfortunately so it's a good compromise for now until we find a better way to do it...

PR #517, opened 2024-02-29
Klippain compatibility with HH v2.5

Benoitone:
try to make klippain compatible with Happy_Hare v2.5
Actually in developpment in happy_hare variables branch

### DIFF
--- a/macros/base/cancel_print.cfg
+++ b/macros/base/cancel_print.cfg
@@ -17,10 +17,16 @@ gcode:
 
     PARK
 
-    {% if klippain_mmu_enabled and mmu_unload_on_cancel_print %}
-        {% if printer.mmu.enabled and printer.mmu.tool|int != -2 %}
+    {% if klippain_mmu_enabled %}
+        {% if printer.mmu.enabled and printer.configfile.settings.mmu.happy_hare_version >= 2.5 %}
+            _MMU_CLEAR_POSITION
+        {% elif printer.mmu.enabled and mmu_unload_on_cancel_print and printer.mmu.tool|int != -2 %}
             # Unload filament and park the MMU
             MMU_EJECT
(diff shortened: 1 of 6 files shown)


Surion79:
could you put this please in a draft state? Or can it be already integrated?

Benoitone:
need some tester before update HH...

Surion79:
A pull request is not required to use/test a branch

Surion79:
Instead of checking for mmu.enabled would it be possible via rename existing to set klippain mmu enabled to false/true if the related function in MMU is used?

Surion79:
i changed the base to develop, hopefully to prevent the issues of last time

Benoitone:
For me it seems to be ok...
The new HH version has been released today...

Surion79:
is it ready for review?

Benoitone:
with this version use with HHv2.5+ no need to have to add entries in printer.cfg
but if they exist they can override HH variables.

Frix-x:
@Benoitone are the changes ok for you?
Especially the startup change as I changed a bit the "if" logic to add the deprecation warning and not sure it's ok like this.

Benoitone:
@Frix-x I rework a little bit the startup. I think it's ok now... can you check?

Benoitone:
> Whats the status?

Working on my side…

Frix-x:
Hello @Benoitone , I'm coming back to the project and will work on merging this branch when you are ready :)
Is there anything that needs to be updated before a full review?

CharlemagneLasse:
@Frix-x @Benoitone does this mean I can't use klippain with a modern MMU (using happy hare v3/v4)? Because I was convinced by various youtube influencers that klippain is the way to go and now after a lot of pain, I need to figure out that this was a bad choice from the beginning? And klippain still really requires a nearly 3 year old, unmaintained version?

PR #522, opened 2024-03-02
optional setting for adaptive primeline

Surion79:
By request:
Added optional variable to enable/disable adaptive primeline. 
Priority:
1. given parameter value
2. variables value (default 1 = enabled)
3. default value (default 1 = enabled)
usage does not change, just a optional variable to setting it by printer has been introduced with this


user_templates/variables.cfg
```diff
@@ -55,6 +55,7 @@ variable_probe_dock_margin_xy: 0, 0
 variable_safe_extruder_temp: 150
 
 ## Prime line parameters
+variable_use_adaptive_primeline: 1 # using the adaptive primeline manages start point and direction automatically
```
Frix-x, line 58: I would change the name to "variable_prime_line_adaptive_mode" to be more aligned with the other variables and would also use True/False to stay aligned with all the other variables.cfg booleans

Frix-x (changes requested):

Surion79:
Waiting for review

arcadien:
👍

Frix-x (approved):

PR #524, opened 2024-03-03
adding option to enable/disable homing in park

Surion79:
added an alternative to prevent homing in park if required.

if m84 is called anywhere and the user forgets it and press park while the bed is full of prints, it does a G28 on the parts.

instead of reverting to the old system with an error, this can be configured via variable with this PR.
default is the "new behaviour" since it is already released


Frix-x:
I will not do this: you are right, I will just revert both macros to the old behavior with the previous alert. It's safer and won't be a big deal for users who call them manually with a G28 before, if they haven't already done so.

Frix-x:
I pushed the commit to main.

PR #525, opened 2024-03-03
added variable for timeout handling

Surion79:
added variable to prevent calling M84 on timeout.
With the G2Z beta, gantry sag is a potenial issue, especially if toolhead is located near the edges.
with the introduced variable, it is now possible to prevent M84 on timeout.

it was copied based on the already existing end print variables, except that i did not added the heater off due to functional safety concerns.

### DIFF
--- a/user_templates/variables.cfg
+++ b/user_templates/variables.cfg
@@ -73,6 +73,8 @@ variable_disable_motors_in_end_print: False
 ## Automatically turn-off heaters in the END_PRINT macro
 variable_turn_off_heaters_in_end_print: True
 
+## Automatically disable motors on timeout
+variable_disable_motors_on_timeout: False
 
 #########################################################
 # Dockable probe variables (if available in the machine)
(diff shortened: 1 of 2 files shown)


Frix-x:
Wouldn't this be a safety issue since a timeout means everything should usually stop to prevent a problem?
Keeping the motors under power seems a bit counterproductive and would add a security flaw IMHO...

user_templates/variables.cfg
```diff
@@ -73,6 +73,8 @@ variable_disable_motors_in_end_print: False
 ## Automatically turn-off heaters in the END_PRINT macro
 variable_turn_off_heaters_in_end_print: True
 
+## Automatically disable motors on timeout
+variable_disable_motors_on_timeout: False
```
Frix-x, line 77: Can you update the default to `True` here? The goal is to align it with the `default(1)` from the macro

Frix-x (changes requested):
Thanks, just a small change about the default value to be able to merge :)

mjonuschat:
I’m going to close this PR for now because feedback hasn't been addressed in a long time.

PR #530, opened 2024-03-04
correct pinout for Fysetc_ERB board

Benoitone:


Frix-x (approved):
Thanks for the fix :)

PR #208, opened 2023-05-17
Add support for LDO Picobilical MCU

ndanyluk:
The LDO Picobilical mod is an umbilical mod that's becoming more and more common as a standalone V0 mod.  Based on the [V0-Umbilical mod](https://github.com/VoronDesign/Voron-Hardware/tree/master/V0-Umbilical), this LDO toolhead & frame board combo adds an RP2040 MCU to the V0 beyond the main microcontroller.  This mod is included with the V0-S1 line of LDO V0 kits (both V0.2-S1 & V0.1-S1).

Original docs can be found at [LDO's Documentation Site](https://docs.ldomotors.com/en/voron/voron01/Picobilical).  Additionally, [schematics](https://github.com/MotorDynamicsLab/LDO-Picobilical/blob/master/Hardware/Frame1.3/Picobilical_A1.3_Schematics.pdf) and [Klipper config](https://github.com/MotorDynamicsLab/LDO-Picobilical/blob/master/Klipper_Configs/ldo-picobilical.cfg) are provided on Github

This PR is still a WIP and needs some guidance on how to add the unique features of this board (2 Neopixel ports that are independent of eachother, a 2 pin filament runout sensor(?) and a small buzzer/speaker/beeper).

PR #533, opened 2024-03-05
Add support for LDO Picobilical

ndanyluk:
Ok, so I suck at git and accidentally screwed up my repo, nuking #208 (and my V0's config) in the process.

I've gone ahead and re-created the config, using the LDO Nitehawk as a reference point.  There are some minor changes from the last known state of #208 (I removed the temp sensor changes and made them local to the config like the Nitehawk has), but the config is mostly identical.

### DIFF
--- a/user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
+++ b/user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
@@ -1,0 +1,65 @@
+#-------------------------------------------------#
+#### LDO Picobilical Frame PCB A1.3 definition ####
+#-------------------------------------------------#
+
+[mcu toolhead]
+##--------------------------------------------------------------------
+serial: /dev/serial/by-id/change-me-to-the-correct-mcu-path
+restart_method: command
+##--------------------------------------------------------------------
+
+# If you want to override the wiring of the LDO Picobilical, keep in mind that this
+# board is defined using the "toolhead" name. So you should use "pin: toolhead:PIN_NAME"
+# in your own overrides.cfg files.
+
+[include config/mcu_definitions/toolhead/LDO_Picobilical_vA1.3.cfg] # Do not remove this line
+[board_pins picobilical_mcu]
+mcu: toolhead
+aliases:
+    X_STOP=MCU_XES ,
+
+    RUNOUT_SENSOR=MCU_FS ,
+
+    E_HEATER=MCU_HE0 , E_TEMPERATURE=MCU_TH0 ,
+
+    PART_FAN=MCU_PCF , E_FAN=MCU_HEF ,
+
+    PB_PCB_TEMP=MCU_CT ,
+
+    # If you want to use both of the independent neopixel ports,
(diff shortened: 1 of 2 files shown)


user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
```diff
@@ -0,0 +1,65 @@
...
+    PART_FAN=MCU_PCF , E_FAN=MCU_HEF ,
+
+    PB_PCB_TEMP=MCU_CT ,
+
+    # If you want to use both of the independent neopixel ports,
+    # comment the below line and uncomment the left/right neopixel line
+    # NOTE: YOU WILL NOT BE ABLE TO USE THE STOCK KLIPPAIN
+    # MACROS WITH THE DUAL NEOPIXEL CONFIGURATION!
+    LIGHT_NEOPIXEL=MCU_NEOPIXEL_1
+    # RIGHT_NEOPIXEL=MCU_NEOPIXEL_1 , LEFT_NEOPIXEL=MCU_NEOPIXEL_2
```
Frix-x, line 34: Can you rename the second one `LIGHT_NEOPIXEL_2` and keep the original `LIGHT_NEOPIXEL` as the first one?

This way this second pin can be used without touching the original Klippain pin and avoid breaking compatibility with the original macros

user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
```diff
@@ -0,0 +1,65 @@
...
+## and having the X endstop plugged to the toolhead MCU
+# [stepper_x]
+# endstop_pin: ^toolhead:X_STOP
+
+[temperature_sensor picobilical_temp]
+sensor_type: CMFB103F3950FANT
+sensor_pin: PB_PCB_TEMP
+pullup_resistor: 2200
+min_temp: 0
+max_temp: 100
```
Frix-x, line 65: Can you add two blank new lines at the end of the file to allow it to be injected to the mcu.cfg user file easily?

Frix-x (changes requested):
Thanks, I think I'm almost ready to merge it. Just a slight change regarding the neopixel pins to avoid breaking compatibility with the other original Klippain files

user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
ndanyluk: whoops, I totally missed this sorry! Will fix now

user_templates/mcu_defaults/toolhead/LDO_Picobilical_vA1.3.cfg
ndanyluk: Done, closing

ndanyluk:
Hey @Frix-x I think this should do it - for some reason Github doesn't want to display the blank 2 lines at the end of the file - is that normal?

Frix-x (approved):

Frix-x:
Thanks, it's ok for me! I'll merge it :)

PR #535, opened 2024-03-05
fix: v0_display toolhead prefix + improve features hardware docs

infowolfe:


docs/features.md
```diff
@@ -23,3 +23,14 @@ Here you can find a list of all the custom features availables in the macros or
...
+If something in your hardware isn't working as expected, inspect the relevant default configuration file for your hardware. For example if your v0 display encoder is rotating in the opposite direction:
+`cd ~/printer_data/config` then `less config/hardware/displays/V0_display.cfg`, copy the relevant portion then add edit to suit in your `overrides.cfg`:
+```
+[display]
+# Set the direction of the encoder wheel
+#   Standard: Right (clockwise) scrolls down or increases values. Left (counter-clockwise scrolls up or decreases values.
+#encoder_pins: ^v0_display:PA3, ^v0_display:PA4
+#   Reversed: Right (clockwise) scrolls up or decreases values. Left (counter-clockwise scrolls down or increases values.
+encoder_pins: ^v0_display:PA4, ^v0_display:PA3
+```
```
Frix-x: I would move and merge this into the overrides documentation here: https://github.com/Frix-x/klippain/blob/main/docs/overrides.md

Then here, you can just add a link with a sentence like: "If some of your own devices are not (yet) supported, Klippain is fully modular and allows you to extend its functionality with your own devices and features by using [overrides](./overrides.md)."

Frix-x (changes requested):
Thanks for the fix! I've just got a slight change request regarding the documentation paragraph you added.

docs/features.md
infowolfe: should be better now, please re-review.

Frix-x (approved):
Thanks it's now ok :)

PR #540, opened 2024-03-09
Update rpi_temp.cfg

EricZimmerman:
Format with _ so a space is in the name, like other temps


EricZimmerman:
compare with [toolhead_mcu_temp.cfg](https://github.com/Frix-x/klippain/blob/EricZimmerman-formatting/config/hardware/temperature_sensors/toolhead_mcu_temp.cfg)

Surion79 (approved):

PR #541, opened 2024-03-09
Neopixel_caselight_enable

Benoitone:
if no toolhead light is selected in printer.cfg and only one of the 2 types of neopixel caselight is included then no lighting will be active in the macros.

### DIFF
--- a/config/hardware/lights/neopixel_caselight.cfg
+++ b/config/hardware/lights/neopixel_caselight.cfg
@@ -1,6 +1,7 @@
 # Neopixel leds used as general printer lights
 
 [gcode_macro _USER_VARIABLES]
+variable_neopixel_leds_enabled = True
 variable_status_leds_caselight_enabled = True
 variable_status_leds_effects_enabled = False
 variable_status_leds_caselight_led_name: "caselight"
--- a/config/hardware/lights/neopixel_caselight_effects.cfg
+++ b/config/hardware/lights/neopixel_caselight_effects.cfg
@@ -2,6 +2,7 @@
 # using LED_effects
 
 [gcode_macro _USER_VARIABLES]
+variable_neopixel_leds_enabled = True
 variable_status_leds_caselight_enabled = True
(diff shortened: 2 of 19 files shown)


Frix-x:
Hello there is also #561 that was open. Can you have a look and tell me which option is better? Or maybe merge both into a global new PR with the best of both?

I'm not using neopixels on my machines so I can't test that much this kind of things...

Benoitone:
> Hello there is also #561 that was open. Can you have a look and tell me which option is better? Or maybe merge both into a global new PR with the best of both?
> 
> I'm not using neopixels on my machines so I can't test that much this kind of things...

Hello, I don't really see the point of adding a variable in variables.cfg...
For me, the modification made solves the problem encountered and seems to me to be entirely compatible with the different possible configurations.

Benoitone:
@Frix-x @mattpolito maybe we can make a common PR with #563 :
```
[gcode_macro _INIT_LEDS]
gcode:
    {% if printer["gcode_macro _USER_VARIABLES"].neopixel_leds_enabled or printer["gcode_macro _USER_VARIABLES"].light_enabled %}
        {% if printer["gcode_macro _USER_VARIABLES"].caselight_on_at_startup|default(False) %}
            {% if printer["gcode_macro _USER_VARIABLES"].light_enabled %}
                LIGHT_ON
            {% endif %}
            {% if printer["gcode_macro _USER_VARIABLES"].neopixel_leds_enabled %}
                STATUS_LEDS COLOR="READY"
            {% endif %}
        {% else %}
            {% if printer["gcode_macro _USER_VARIABLES"].light_enabled %}
                LIGHT_OFF
            {% endif %}
            {% if printer["gcode_macro _USER_VARIABLES"].neopixel_leds_enabled %}
                STATUS_LEDS COLOR="OFF"
            {% endif %}
        {% endif %}
    {% endif %}
```

mjonuschat:
Fixed in PR #756

PR #556, opened 2024-03-14
KlipperScreen_restart

Benoitone:
After an update of klipper or klippain or happy_hare for example klipperscreen does not restart.
This PR allows the restart of the klipperscreen service (if it exists) during a Restart.
Not knowing if this could be interesting for other services, it is possible to complete the script with a list of services:
ex: `services = ['KlipperScreen', 'service1']`


Surion79:
We shouldn't program something that has moonraker update entries by itself using the configuration option "managed_services".
This is referenced in the moonraker documentation [https://moonraker.readthedocs.io/en/latest/configuration/](url) and in the KlipperScreen documentation [.](https://klipperscreen.readthedocs.io/en/latest/Installation/#moonraker-configuration )

PR #561, opened 2024-03-19
Option to use RGB caselights as status LEDs

teookie:
Modified calls to gcode macro STATUS_LED to also execute if user variable status_leds_caselight_enabled is true.

Also added variable status_leds_caselight_enabled to user variables file. If user sets this to true, then caselights will function as status LEDs. If set to false, then caselights will illuminate to their initial colors as called out in [neopixel caselight], but won't function as status LEDs.

(This is my first ever pull request on github - I hope I did it right.)

### DIFF
--- a/macros/base/cancel_print.cfg
+++ b/macros/base/cancel_print.cfg
@@ -10,6 +10,7 @@ gcode:
     {% set filter_enabled = printer["gcode_macro _USER_VARIABLES"].filter_enabled %}
     {% set light_enabled = printer["gcode_macro _USER_VARIABLES"].light_enabled %}
     {% set status_leds_enabled = printer["gcode_macro _USER_VARIABLES"].status_leds_enabled %}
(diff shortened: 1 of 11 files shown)


Frix-x:
Hello thanks for the PR! There is also #541 that was open. Can you have a look and see with Benoit which option is better and it could be made into a new PR that group both of the ideas?

Benoitone:
> Hello thanks for the PR! There is also #541 that was open. Can you have a look and see with Benoit which option is better and it could be made into a new PR that group both of the ideas?

@teookie I don't understand exactly the utility of `variable_status_leds_caselight_enabled` in `variables.cfg` because if you have toolhead status_led also included, caselight (not Fcob) use also status leds macros... So I think if set to `false/true` nothing changes...

mjonuschat:
Fixed in PR #756

PR #569, opened 2024-03-26
Nozzle cleaning parameters update

ksummers92:
This update is to add parameters to the nozzle brushing macro to allow for further customization. The impetus for this update is explained in the "brush_center_offset.md" file. Values have been tested to work.


Surion79:
Please make sure that the new variables have a meaningful default value in the macros you changed. Klippain Updaters do not get new variables in the variables.cfg integrated. Current state this update would break Klippain for users of nozzle brushing functionality.

Frix-x (approved):
I'm mostly ok with this. I'll merge and will reference the docs when the website will be up (I want to work on a proper doc website)

PR #570, opened 2024-03-26
Update to remove spaces from pin definitions after = across multiple boards.

o1hitman1o:
Remove extra space characters from pin definitions. Not sure if it is an issue but cleans up the look.


Surion79 (approved):
looks good to me

Frix-x (approved):
Thanks for catching these :)

PR #579, opened 2024-04-16
Message update to show "Available RAM" rather than "Used RAM"

cjohnson-phys:
Just a small STDOUT update to show "Available RAM" instead of "Used RAM". The measurement made in the code shows Available RAM but the message says "Used RAM".

Can be a little disconcerting if you don't double check :)


Surion79 (approved):
thanks for the correction :)

Frix-x (approved):

PR #582, opened 2024-04-18
added a fix for filament sensor activation for material as default

Surion79:
in case the filament_sensor setting is missing in the material, it is defaulted to 1.
I used this approach to remove a variable and use a single variable to define on start_print if the sensor is enabled or not.


Frix-x (approved):
Thanks, I'll merge

PR #597, opened 2024-06-04
added start macro parameter and handling to skip bed mesh

HerrStinson:
Useful if you want to print multi color first layer, after the first color is printed the bed is still level but bed mesh will fail obviously. Quick and dirty fix for this is to add SKIP_BEDMESH=1 to macro in slicer.

### DIFF
--- a/config/software/shaketune.cfg
+++ b/config/software/shaketune.cfg
@@ -2,6 +2,6 @@
 # to work correctly in the full Klippain environement (within the scripts folder).
 
 [gcode_shell_command shaketune]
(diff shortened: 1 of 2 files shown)


Frix-x:
Thanks for the PR!
To be able to merge it, can you also add an entry in the documentation to specify the new parameter? Like here: https://github.com/Frix-x/klippain/blob/main/docs/configuration.md#custom-print-start-gcode

HerrStinson:
Will do, but as I am currently abroad it will take a couple of days.

Frix-x:
No problem, take your time :)

mjonuschat:

PR #606, opened 2024-06-20
added function for pause at (next) layer

Surion79:


macros/base/cancel_print.cfg
```diff
@@ -77,4 +77,8 @@ gcode:
         SET_FILAMENT_SENSOR SENSOR="runout_sensor" ENABLE=1
     {% endif %}
     
+    # clear pause_next_layer and pause_at_layer as preparation for next print
+    SET_PAUSE_NEXT_LAYER ENABLE=0
```
Frix-x, line 81: This can probably be added also in the standard END_PRINT macro in case there was a wrong config or something happened to be sure it's reset correctly

Surion79: done

Frix-x (approved):

PR #607, opened 2024-06-21
simple cartographer implementation

Surion79:
@Frix-x : this is the beacon similar "simple" implementation


Surion79:
double found

PR #619, opened 2024-07-10
Allow homing commands to be over-ridden

kerinin:
Currently, the only way to customize the homing logic is to define a `[homing_override]` override, which is overkill for some applications. This creates `_KLIPAIN_HOME_XXX` macros that are called in the existing `homing_override` and can be overridden to customize homing behavior in a more targeted way.

I made these changes in because I use sensorless homing on one axis only, and I wanted to create a custom macro similar to https://klipper.discourse.group/t/sensorless-homing-macro-suggested-in-klipper-docs/896 that allows me to home twice with a pause between for StallGuard to reset.

### DIFF
--- a/macros/base/homing/homing_override.cfg
+++ b/macros/base/homing/homing_override.cfg
@@ -1,4 +1,26 @@
 # Homing Override for dockable probe
+[gcode_macro _KLIPAIN_HOME_X]
+gcode:
+    G28 X0
+    G1 X{params.X_POSITION_ENDSTOP|float + params.X_HOMING_BACKOFF|float} F{params.HOMING_TRAVEL_SPEED|float}
+
+[gcode_macro _KLIPAIN_HOME_Y]
+gcode:
+    G28 Y0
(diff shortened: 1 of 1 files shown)


Surion79:
for the beginning, i suggest to write the name of this repo correctly in the functions. there is a letter missing

Frix-x:
Hello, thanks for the PR!

I'm pretty open for this change as it will indeed allow to split and simplify the logic :)
However, let's keep it simple:
  1. Can you change the names to `_HOME_X`, `_HOME_Y` and `_HOME_Z`? No need to add Klippain name here
  2. Since we are in the homing process, there's no macro parameters beside X, Y, Z to specify which axes to home. So in order to keep the `_HOME_...` macros independent, could you avoid passing parameters and directly read from `_USER_VARIABLES` and printer object? This will simplify even further and will avoid mistakes.

Then I think I'll be able to merge it :)

mjonuschat:

Issue #587, opened 2024-05-07
Manta M8P v2.0 FAN3 pin incorrect


AndyL069:
### Klippain branch

- [X] I confirm using the main branch

### Version

v4.3.1-11-g4ff6f5b4

### Describe the bug and expected behavior

The pin assignment for MCU_FAN3 (Exhaust filter) is wrong.
It is MCU_FAN3=PA8, it should be MCU_FAN3=PF8

https://github.com/bigtreetech/Manta-M8P/blob/master/V2.0/BIGTREETECH%20MANTA%20M8P%20V2.0%20User%20Manual.pdf

AndyL069:
![image](https://github.com/user-attachments/assets/34182cc9-a2c6-4ee2-806b-e5623ad2e481)

AndyL069:
I created a pull request for this issue: https://github.com/Frix-x/klippain/pull/628

PR #621, opened 2024-07-12
Update BTT_Manta_M8P_v2.0.cfg

smwoodward:
Fixed MCU_FAN3 pin number.


AndyL069:
fixes #587

AndyL069 (approved):
Change is correct

excieve:
This is still relevant and shouldn't be stale, the fan pinout in `main` is currently incorrect for this board.

Frix-x:
Everything is alright. It was merged from another PR that was fixing the same thing :)
(I didn't see this one was also here)

So I'll close it as it's solved

PR #627, opened 2024-07-22
Add 2 TMC motor/stepper Z axis support

stablestud:
This comes in handy for dual lead screw systems, which are wired to seperate drivers.
Also this aids to make use of z_tilt with 2 Z axis TMC steppers


stablestud:
Closing as duplicate

PR #631, opened 2024-07-25
Add 2 TMC motor/stepper Z axis support

stablestud:
This comes in handy for dual lead screw systems, which are wired to seperate drivers.
Also this aid to make use of z_tilt with 2 Z axis TMC steppers


stablestud:
Closing this I'm not using klippain because of its complexity. I'm using [klipper-macros](https://github.com/jschuh/klipper-macros) instead

PR #632, opened 2024-07-30
feature(tap) Added chaoticlab CNC tap settings.

jackmoxley:
Adding the choatic lab cnc probe, which has a higher threshold as it is more prone to failure on gantry checking and height maps.

It used the base voron_tap, so all the macros and pin settings remain.

### DIFF
--- a/config/hardware/probes/chaoticlab-cnc-tap-v2.cfg
+++ b/config/hardware/probes/chaoticlab-cnc-tap-v2.cfg
@@ -1,0 +1,12 @@
+# Chaotic Lab CNC Tap is a little less sensitive, and needs it's own settings.
+# It is based on the default voron_tap config
+# @see (https://github.com/Chaoticlab/CNC-Tap-for-Voron/blob/master/Manual/CNC%20Tap%20V2.cfg)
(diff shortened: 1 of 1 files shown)


EricZimmerman:
Why is this needed vs just straight up overriding the [probe] block?

jackmoxley:
> Why is this needed vs just straight up overriding the [probe] block?

I did exactly that, but in the nature of providing out the box support for everyone that uses these mods, I thought it would be helpful.

You could say the same thing about anny klippain config file.

Frix-x (changes requested):
Hey, thanks for the PR!

I'm all for it. It's part of the work I wanted to do to make more probes easily available without having to put the offsets in the overrides every time. So this might be a good candidate for a first file like this.

Eric, the goal would be to be able to select with Klicky, Euclid, etc... instead of "dockable", since not all dockable probes are similar. But they will all reference this dockable probe file under the hood with the added customizations.

However, @jackmoxley can you add the entry in the printer.cfg file to be able to select it? Thanks!

mjonuschat:

PR #641, opened 2024-08-26
Make step to add pressure in the nozzle configuable

tehniemer:
Add user configurable variable for the extrude distance in the step adding pressure in the nozzle in the prime line macro


tehniemer:

tehniemer:
Added functionality into #642

PR #642, opened 2024-08-27
Add filament retract and unretract user variables for use in base and helper macros

tehniemer:
Add user configurable variable for filament retract in end and cancel print macros.


tehniemer:

Frix-x:
Thanks for the PR!

I want to merge this. But as stated in #658 this need to be linked to the loading during a purge at the beginning of a print. Indeed, what you retract at the end of a print should be inserted again at the beginning of the next one (at least it was designed like that).

Can you please complete the PR with this additional need?

wight554:
In my testing I noticed that it's better to push a little bit more than retracted, like 10-20% more 
I retract 30 and push 35, otherwise half of my purge line is somewhat empty. However would be better if anyone else shares their experience

tehniemer:
I just added an unretract to the prime line macro equal to the retract in the cancel and end print macros. Please let me know if that's good enough or if you would like an additional user variable for the extra push as mentioned by @wight554

Frix-x:
Yes, that's right, I think having a percentage factor combined to this could be a good idea.
Like by default 15% more: it doesn't cost much to do this and should be pretty harmless.

tehniemer:
OK, I've added #641 and #645 into this one and created a second variable for the nozzle prime.

macros/helpers/nozzle_cleaning.cfg
```diff
@@ -120,7 +121,7 @@ gcode:
         # Retract
         G92 E0
         G1 E-1.7 F2100
-        G1 E-18.3 F150
+        G1 E-({RETRACT}-1.7) F150
```
Frix-x, line 124: Can you replace it by like 10% like on the following suggestion? It should be pretty equivalent but also more compatible with any value of RETRACT.

```suggestion
        G1 E-({RETRACT}-{RETRACT*0.1}) F150
```

user_templates/variables.cfg
```diff
@@ -61,6 +61,12 @@ variable_prime_line_height: 0.6 # mm, used for actual cross section computation
 variable_prime_line_margin: 5  # distance of purge line from fl_size rectangle
 variable_prime_line_wipe: False # enable a wipe of the nozzle after completing the prime line
 
+## Length of filament to retract to reduce oozing at end of print and during purge sequence, if enabled (in mm)
+variable_retract_length: 20
+
+## Length of filament to pre-prime the nozzle before start print (in mm)
+variable_unretract_length: 23
```
Frix-x, line 68: Maybe for it to be more clear and show that both variables are linked, it could be made like this.

```suggestion
## Length of filament to retract/retract to disengage it from the heatbreak.
## This is to reduce final print oozing, but also to prevent heatcreep, clogging, and allow cold filament change between prints.
variable_retract_length: 20
variable_unretract_length: 23
```

Frix-x (changes requested):
I think we are near a merge. Just a few changes on my side, but nothing very problematic :)

tehniemer:
I'm good with both of those.

Frix-x (approved):

Frix-x:
Ok I'm good with it now and will merge! Thank you :)

PR #645, opened 2024-08-31
Make retract user configurable in nozzle cleaning

tehniemer:
Added user configurable variable for final retract in the nozzle cleaning macro.


tehniemer:

tehniemer:
Added functionality into #642

PR #652, opened 2024-10-04
Update start_print.cfg

ksummers92:
Update CHAMBER param to CHAMBER__TEMP to match other temp param names


Elias23:
If you change the name, you probably should also update the documentation to reflect your change. Maybe consider some kind of fallback logic, so this change doesn't break the current installations, just an idea.

Frix-x:
Yes, this PR is a bit complicated to merge as it would be a breaking change for almost everyone... So I prefer to keep current naming if possible...

PR #654, opened 2024-10-10
Mini12864 beeper change due to klipper change

rongith:
Klipper Configuration Changes: "20240123: The output_pin SET_PIN CYCLE_TIME parameter has been removed. Use the new pwm_cycle_time module if it is necessary to dynamically change a pwm pin's cycle time."

Edit: It will break compatibility with older klipper versions.


rongith:
I'm a still bad at git. This PR concern only the commit [2d6fe98db3456d5e6402e789f434c4c8f9b9371c](https://github.com/Frix-x/klippain/pull/654/commits/2d6fe98db3456d5e6402e789f434c4c8f9b9371c). I need to rebranch.

PR #658, opened 2024-10-17
Add a variable distance for end_print extruder pull back.

rongith:
I like to have the end_print pull back distance a variable.
Why: I have different toolheads with different hot zone lengths and I like the filament pulled in the cold zone according to the toolhead.

### DIFF
--- a/macros/base/end_print.cfg
+++ b/macros/base/end_print.cfg
@@ -25,7 +25,7 @@ gcode:
     {% elif printer.extruder.can_extrude %}
         # pull back the filament a little bit
         G92 E0
-        G1 E-10 F2100
+        G1 E{extruder_pull_back_distance_end_print} F2100
     {% endif %}
 
     {% if turn_off_heaters_in_end_print %}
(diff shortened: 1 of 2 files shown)


wight554:
@rongith thanks for raising PR
you might want to remove hardcode of 18mm here as well
https://github.com/Frix-x/klippain/blob/main/macros/helpers/prime_line.cfg#L113

In my issue I suggested retracting same amount as you pre-extrude in the following line
Might be good idea to create another option for this case as well

rongith:
@wight554 , I may do an other PR for that. Perhaps a better practice.
Here you want to keep a relation between the "pressure in the nozzle" before prime in PRIMELINE and the previous pull in END_PRINT?

wight554:
I honestly feel like at start you should push amount you retracted in the end, so purging is more predictable. But we can configure both separately

Frix-x:
Thanks for the PR!

This is a duplicate of #642, so I'll close it in favor of the other one. But on paper it's a good idea and I'll merge the feature. I just want to add, as commented by @wight554, that this parameter is related to the loading of the filament during the purge, before a print. In fact, what you unload at the end of a print should be reloaded at the beginning of the next print (at least it was designed that way).

PR #662, opened 2024-10-21
Update homing_override.cfg with a M400 to ensure the nozzle is lifted

walndyn:
Add a M400 command after the z hop gcode to ensure the nozzle is lifted before anything else is allowed to happen to prevent nozzle crashes. This happend while testing out my new Cartographer3D eddy current bed probe where somehow the cartographer klipper plugin executed a move to the homing position before the nozzle was lifted from the bed surface an scratched it.


Frix-x (approved):

walndyn:
Thank you for merging and for maintaining this repo :)

Issue #593, opened 2024-06-01
Make end print retract amount a variable


wight554:
### Describe the feature or hardware support you'd like

Currently primeline purges 18mm and end_print retracts 10mm which creates a blob at the start of primeline. In my setup my fan duct keeps grabbing that blob when purgeline is in X direction which melts the duct after some abs prints

Similar to KAMP's https://github.com/kyleisah/Klipper-Adaptive-Meshing-Purging/blob/main/Configuration/KAMP_Settings.cfg#L26 we can benefit even more from having variable accessible both from prime line and end_print

Lines of code to change:
https://github.com/Frix-x/klippain/blob/main/macros/helpers/prime_line.cfg#L113
https://github.com/Frix-x/klippain/blob/main/macros/base/end_print.cfg#L28

Surion79:
so these are practically two changes? One for end print and one for prime line?

wight554:
@Surion79 yeah but they should use same variable (that has to be created as well). The idea is to retract same amount that'll be pushed before purge

rongith:
To the OP: I created a PR for that: https://github.com/Frix-x/klippain/pull/658
It seems we can't reopen an old issue anymore.

PR #665, opened 2024-10-28
Add a variable distance for prime_line nozzle pressure push.

rongith:
We don't know for sure how far the filament is in toolhead before prime_line. For the moment the prime_line push is hard coded to 18. I propose to set it as a variable then one want to adjust it.
Request from @wight554 here: https://github.com/Frix-x/klippain/issues/593


Frix-x:
Hello, thanks for the PR!

But this is a duplicate of #642 that I just merged in Klippain. So this issue is solved now and I'll close it :)

PR #672, opened 2024-11-06
feat: support 2 motor z axis

ammmze:
Support for printers with dual z axis (such as my Two Trees Sapphire Plus)

### DIFF
--- a/config/hardware/axis/Z/default_wiring_2M.cfg
+++ b/config/hardware/axis/Z/default_wiring_2M.cfg
@@ -1,0 +1,9 @@
+# In this file we include the first motor
+[include default_wiring_1M.cfg]
+
+# And then, add a 2nd motor...
+[stepper_z1]
+step_pin: Z1_STEP
+dir_pin: Z1_DIR
+enable_pin: !Z1_ENABLE
+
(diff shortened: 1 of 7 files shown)


config/hardware/axis/Z/default_wiring_2M.cfg
```diff
@@ -0,0 +1,9 @@
+# In this file we include the first motor
+[include default_wiring_1M.cfg]
+
+# And then, add a 2nd
```
NanoMetere, line 4: "And then, add a 2nd motor."

NanoMetere (commented):
There's a little comment error in default_wiring_2M.cfg, but codes are complete, and it'll work fine.

config/hardware/axis/Z/default_wiring_2M.cfg
ammmze: 👍🏻 Thank you! I've updated the comment to be consistent with the other stepper files.

NanoMetere (approved):
Great. All files are reviewed and should work fine for everyone.

max-scopp:
@Frix-x is there anything blocking this? I’m migrating from my custom config solution to Klippain and would’ve loved to see this as a default option.

mjonuschat:
Shouldn't there at least be some template in `config/hardware/axis/Z` for a printer with two Z axis? This seems to be just the low level wiring. Seems like the Sapphire Plus would be a candidate?

mjonuschat:
Closing this as stale/incomplete

PR #676, opened 2024-11-18
Update end_print.cfg

LegionPleingaz:
Move Motor OFF at the end of the script due to gantry drop with a hot HE. This is a problem with a full plate

### DIFF
--- a/macros/base/end_print.cfg
+++ b/macros/base/end_print.cfg
@@ -45,10 +45,6 @@ gcode:
         BED_MESH_CLEAR
     {% endif %}
 
-    {% if disable_motors_in_end_print %}
-        M84
-    {% endif %}
-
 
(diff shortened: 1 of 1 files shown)


Benoitone:
> Move Motor OFF at the end of the script due to gantry drop with a hot HE. This is a problem with a full plate

It's a variable parameter in variable.cfg
Don't you have True as variable_disable_motors_in_end_print variable?
https://github.com/Frix-x/klippain/blob/a30308e5316e50191e21f660d7fdcee5681dcdf6/user_templates/variables.cfg#L68

LegionPleingaz:
`## Automatically disable motors in the END_PRINT macro
variable_disable_motors_in_end_print: False`

Variable is at "false"
i didn't know it btw, so something is not working. My fix may not be the solution

PR #684, opened 2025-01-08
Add Cartographer3D V3 with Touch configuration

typetetris:
This adds configuration for a Cartographer 3D with Touch enabled and an adxl accelerometer.

Please give it a thorough review, as its my first work on such stuff, which I didn't do only for myself.

Reasons for some of the decisions I made:

We need to be able to override some settings for
`[bed_mesh]` so the include in `user_templates/printer.cfg` is placed after the bed mesh definitions and not
in the probe section.

Cartographer3D V3 doesn't touch the bed with the nozzle except for the `CARTOGRAPHER_TOUCH` command. Therefore we need to pay attention to the nozzle temperature for that. But not for all the other actions like
homing or quad gantry leveling or some such.

The documentation for Cartographer3D advises against an adaptive bed mesh and the full bed mesh will
be fast enough in most cases, so we deactivated
the adaptive bed mesh if a Cartographer3D probe
is used. Maybe this should be configurable? Tell me, and I will alter it.

Also, what do you think about `variable_cartographer_touch_max_probing_temp` maybe I should simply use
`variable_safe_extruder_temp` instead?

This would make it easier and more backwards compatible. But for now I took inspiration
from `variable_tap_max_probing_temp`.

### DIFF
--- a/config/hardware/accelerometers/cartographer_v3_accelerometer.cfg
+++ b/config/hardware/accelerometers/cartographer_v3_accelerometer.cfg
@@ -1,0 +1,6 @@
+[include generics/adxl345_hardware_spi1.cfg]
+
+# You need to include the section for the cartographer in your printer.cfg and include cartographer_virtual.cfg
+# and set the correct serial for this to work.
+[adxl345]
+cs_pin: scanner:PA3
--- a/config/hardware/probes/cartographer_v3_virtual.cfg
+++ b/config/hardware/probes/cartographer_v3_virtual.cfg
@@ -1,0 +1,48 @@
+# This probe type is for a cartographer probe used directly as a virtual Z endstop
+# rather than with an existing physical endstop. To use this configuration,
+# you will need to manually add the cartographer Klipper plugin!
+
+[gcode_macro _USER_VARIABLES]
+variable_probe_type_enabled: "cartographer_touch"
+variable_startprint_actions: "bed_soak", "extruder_preheating", "chamber_soak", "tilt_calib", "z_offset", "bedmesh", "clean", "cartographer_touch", "extruder_heating", "purge", "clean", "primeline"
+variable_homing_zhop: 10
+gcode:
+
+# Cartographer 3D probe definition also include the probe management macro directly from here
+[include ../../../macros/base/probing/generic_probe.cfg]
+
+[scanner]
+mcu: scanner
+#   Offsets are measured from the centre of your coil, to the tip of your nozzle
+#   on a level axis. It is vital that this is accurate.
+x_offset: 0
+#    adjust for your cartographers offset from nozzle to middle of coil
+y_offset: 15
+#    adjust for your cartographers offset from nozzle to middle of coil
+backlash_comp: 0.5
+#   Backlash compensation distance for removing Z backlash before measuring
+#   the sensor response.
+sensor: cartographer
+#    this must be set as cartographer unless using IDM etc.
+sensor_alt: carto
+#    alternate name to call commands. CARTO_TOUCH etc
+mesh_runs: 2
+#    Number of passes to make during mesh scan.
+
+[bed_mesh]
+speed: 200
+#    movement speed of toolhead during bed mesh
+horizontal_move_z: 5
+#    height of scanner during bed mesh scan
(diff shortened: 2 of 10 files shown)


typetetris:
> So the first thing would be to use the ACTIVATE/DEACTIVATE_PROBE to manage the temperature limit, etc... and remove the cartographer_touch action that will not be needed anymore as it can be a simple z_offset instead.

Regarding using the `z_offset` action: At the moment I use that to home the z axis again, after doing the `tilt_calib`, which is recommended by the Cartographer documentation.

Should I include rehoming the z axis in the `tilt_calib` action if the `probe_type_enabled` is `cartographer_touch` or would you suggest a different approach?

Frix-x:
> Should I include rehoming the z axis in the `tilt_calib` action if the `probe_type_enabled` is `cartographer_touch` or would you suggest a different approach?

No, this is not necessary. In fact, it is necessary for all probes to do this: after a QGL, you have to do a Z home to take a new, more correct height reference. But in Klippain, to avoid wasting time and doing too many Z homeings during the START_PRINT, especially for probes that need to have a lower temperature like TAP, the start print sequences are made by taking this into account automatically and the Z homeings are done most of the time only once, just before taking the final Z offset. So there's no need to add specific things, as it's already taken into account later :)

elpopo-eng:
@typetetris We've done something a litle bit different and more transparent in [klippain-chocolate](https://github.com/elpopo-eng/klippain-chocolate). We use carto like a voron tap-probe
You can take a look [here](https://github.com/elpopo-eng/klippain-chocolate/blob/404de828fd29cd3763b05f137f905279532bfb90/config/hardware/probes/cartographer_survey.cfg#L37) and [here](https://github.com/elpopo-eng/klippain-chocolate/blob/404de828fd29cd3763b05f137f905279532bfb90/macros/base/start_print.cfg#L395)

Frix-x:
Thanks Elpopo, that's indeed a pretty good and clean way to do it. Do you
want to PR it to Klippain for integration?

DisDis:
It works nice. thx. @typetetris 
When will it merge into master?
How can I help?

Sidirius:
> It works nice. thx. @typetetris When will it merge into master? How can I help?

Second this, anything I / we can do to help for it to merge into master?

ubrider:
Any updates ? 
receiving mine soon !

Cicatr1x:
@elpopo-eng could you create a PR with your Changes, so @Frix-x can merge it?

elpopo-eng:
Sorry, but it's easier to switch to klippain-chocolat.

Cicatr1x:
> Sorry, but it's easier to switch to klippain-chocolat.

That would make keeping on track with Frix-x`s changes harder.

EuroAfrican (approved):

Cicatr1x:
@EuroAfrican without really beeing invested and understanding the code a review is pretty much worthless, especially when you dont have the needed privileges 😉

Cicatr1x:
@elpopo-eng where do you specify the cartographers "canbus_uuid" in your repo config?

elpopo-eng:
> @elpopo-eng where do you specify the cartographers "canbus_uuid" in your repo config?

https://github.com/elpopo-eng/klippain-chocolate/blob/develop/config%2Fhardware%2Fprobes%2Fbeacon_virtual.cfg#L5

Cicatr1x:
> > @elpopo-eng where do you specify the cartographers "canbus_uuid" in your repo config?
> 
> https://github.com/elpopo-eng/klippain-chocolate/blob/develop/config%2Fhardware%2Fprobes%2Fbeacon_virtual.cfg#L5

Even when i dont include the beacon because i included Cartographer? I Just added this in my printer.cfg.

[mcu Scanner]
can_uuid: mycanuuid

ErdbeermarmeladebrotmitHonig:
Hi,
I opened a pull request here: #743 
This also uses the newer macros comming with the cartographer firmware.

mjonuschat:
Implemented in the Z probe framework, confirmed working after 6ea2917db10957f2068b6ac7b7b133ef2c05d36d.

Closing this out as duplicate.

PR #687, opened 2025-01-09
Add an option for a user controllable exhaust fan.

Elias23:
I use an exhaust fan coupled with a hepa filter and hose to keep my chamber at a slight negative pressure and to vent the chamber out a window.

This PR would add this option to klippain with variables and the option to configure the exhaust on a per material basis.

### DIFF
--- a/config/hardware/filters/controllable_exhaust_filter.cfg
+++ b/config/hardware/filters/controllable_exhaust_filter.cfg
@@ -1,5 +1,4 @@
 [gcode_macro _USER_VARIABLES]
-variable_startprint_actions: "bed_soak", "extruder_preheating", "chamber_soak", "clean", "tilt_calib", "z_offset", "custom1", "bedmesh", "extruder_heating", "purge", "clean", "primeline"
 variable_exhaust_enabled: True
 variable_exhaust_name: "exhaust"
 gcode: # do not remove this line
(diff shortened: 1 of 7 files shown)


mjonuschat:
Hey, finally catching up with some of the older PRs, there are some issues with this one:

- **Variable name mismatch → runtime error.** `variables.cfg` defines `exhaust_filter_default_time_on_end_print` but cancel/end macros reference `filter_exhaust_default_time_on_end_print` (swapped prefix). This will crash.
- **`FILTER_TIME` param reuse.** Both cancel and end print pass `params.FILTER_TIME` as the exhaust duration default — meaning a `FILTER_TIME=300` intended for the recirculating filter also sets the exhaust timer. Needs its own param.
- **Indentation bug** in the cancel/end print exhaust blocks.
- **No `initial_duration: 0`** on `_STOP_EXHAUST_DELAYED`.
- **No material template updates** — `material.exhaust_speed` is referenced but never documented or added to any material profile.
- **Architecturally, this duplicates the entire filter subsystem.** A cleaner path would be generalizing filter management to support multiple named instances.

Elias23:
Thank you for the feedback. Do you see value in adding this feature, if yes I would rework the pull request. I'm quite sure this branch is not in sync anymore with what I'm running on my printer.

For the last point if you have an example of a system in klippain that works in that generalized fashion I can try to apply that architecture to the filter system.

mjonuschat:
Personally I've blocked off all the exhausts in my printers with a blanking plate, but I can see that people prefer to use the exhaust and vent to the outside.
I think if we can find a way to not duplicate a large chunk of the fan/filter code this would be a good addition. Could also be a two step PR - start with the controllable fan first, then integrate the materials support.

As to a system that generalizes: The probe framework for Beacon/Carto/PZ/Bltouch is probably a good but also quite involved example. I'll think about how I would approach the fan implementation and see if I can come up with suggestions.

mjonuschat:
Looking good! A few things from reviewing the current diff:

- FILTER_TIME param reuse: 
   cancel and end print both fall back to params.FILTER_TIME for the exhaust duration. A FILTER_TIME=300 meant for the recirculating filter will set the exhaust timer too. Needs its own param, maybe EXHAUST_TIME?
- off_below: 
  Lets drop this to keep compatibility with Klipper and Kalico - a user can easily add this to the `overrides.cfg` if needed for their fan model.
- Hardcoded fan name in cancel/end print: 
   printer['fan_generic exhaust'].speed > 0 bypasses the exhaust_name variable that START_EXHAUST/STOP_EXHAUST correctly use.

PR #690, opened 2025-01-17
Create PZ Probe.cfg

ibrahimbaig4:
Add PZ probe in Klippain


mjonuschat:
Thanks for your contribution. We've recently added a new framework for virtual z endstop probes and I have added the Revo PZ probe there, based on your work. I'm closing out this PR in favor of the new framework (#767)

PR #695, opened 2025-01-28
Fix: Main MCU default should run printer itself.

NanoMetere:
just building a v0 with Manta E3EZ with no toolhead board and found out default template don't have PART_FAN and E_FAN so it cant run with default template.

tested on Manta E3EZ but not Manta M5P. but should work fine.
this changes are for newcomers and no affect to current users.


NanoMetere (commented):
base.conf is just my mistake committed on main. file reverted to original.

PR #698, opened 2025-02-08
Update LDO_Leviathan_v1.2.cfg with pullup

Jothoreptile:
Add pullup to match real temperature (20C of difference with and without the pullup for 70 degrees ask in klipper)


mjonuschat:
Closed by dbfe109 - added the pullup resistors to the user template to keep things consistent

PR #717, opened 2025-08-14
Pinout fix for Leviathan MCU config - Board v1.2 and v1.3 pinout compatible

frostyllama:
LDO_Leviathan_v1.2.cfg Renamed ..._v1.x.cfg - pinout the same for v1.2 and v1.3 boards. Controller update on v1.3 - may need additional testing.

Fixed Fan0 Tacho (PB0 > PB8) and EXP2 SPI_CS (PE4 > PA4) from official LDO Leviathan manuals v1.2 and v1.3 - both crosschecked.

Ref: https://github.com/MotorDynamicsLab/Leviathan/tree/master/Manual


config/mcu_definitions/main/LDO_Leviathan_v1.x.cfg
mjonuschat, line 1: Renaming this file will break all installations that have a user config including the v1.2 named files (older installations). The user_template also still refers to the v1.2 file.

A workable solution should be to rename the user template and leave this file as v1.2 with a comment in the header that it also applies to v1.3

mjonuschat (changes requested):

mjonuschat:
Closed through #751

PR #740, opened 2025-12-12
Ldo nitehawk 36 rev c+

tehniemer:
Add configuration for this board, it differs slightly from the SB version.


neptunus0001:
This is the same as https://github.com/Frix-x/klippain/pull/738

PR #741, opened 2025-12-12
Off below to min power

tehniemer:
Created a kalico specific version of the nevermore filter config as `min_power` is being depreciated for `off_below` sometime in the future.


mjonuschat:
"Fixed" by removing the min_power line for the bed fans config - having extra config files for every little bit where Kalico decides to operate differently from Klipper isn't feasible long term. off_below/min_power can be added to `overrides.cfg`.

Closed by commit 5dc59c39887a6d265ab44a5425e3534e5e0079b3

PR #742, opened 2025-12-27
Add nitehawk 36

Agentspades:
Add nitehawk-36 toolboard to MCU definitions


TheKas:
@Frix-x it would be nice, that this version of nitehawk 36 would be supported in klippain.

neptunus0001:
This is the same as https://github.com/Frix-x/klippain/pull/738

PR #743, opened 2026-01-01
added support for Cartographer 3D

ErdbeermarmeladebrotmitHonig:
Hi,
I just installed my Cartographer V4 into my Voron 2.4 and added the corresponding configuration files. I just tested them using some of the code mentioned in #684 and [klippain-chocolate](https://github.com/elpopo-eng/klippain-chocolate). I also added the config for the Cartographer V3, however this is untested for now as I do not own a Cartographer V3.

### DIFF
--- a/config/hardware/accelerometers/cartographer_adxl_accelerometer.cfg
+++ b/config/hardware/accelerometers/cartographer_adxl_accelerometer.cfg
@@ -1,0 +1,15 @@
+# Only for cartographer V4 with adxl
+# You need to include cartographer_survey.cfg and set the correct serial/canid in your overrides for this to work
+[adxl345] 
+cs_pin: scanner:ADXL_CS
+spi_bus: spi1
+
+[resonance_tester]
+accel_chip: adxl345
+probe_points:
+    -1,-1,-1
+
+# Include the IS calibration macros to unlock them when
+# an accelerometer is installed on the machine
+[include ../../../macros/helpers/resonance_override.cfg]
+[include ../../../scripts/K-ShakeTune/K-SnT_*.cfg]
--- a/config/hardware/probes/cartographer_survey.cfg
+++ b/config/hardware/probes/cartographer_survey.cfg
@@ -1,0 +1,33 @@
+# This probe type is for the Cartographer 3D probe used directly as a virtual Z endstop
+# rather than with an existing physical endstop. To use this configuration,
+# you will need to manually add the Cartographer 3D Klipper plugin!
+
+## Inside your mcu.cfg file, you need to add the cartographer as mcu like this:
+# [mcu scanner]
+# serial: change-me-to-the-correct-canbus-id
+# canbus_uuid: /dev/serial/by-id/change-me-to-the-correct-mcu-path
+# restart_method: command # Not needed for probes using CAN.
+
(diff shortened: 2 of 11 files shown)


TheKas:
@Frix-x it would be nice, that version V4 of Cartographer 3D would be supported in klippain.

config/hardware/accelerometers/cartographer_adxl_accelerometer.cfg
```diff
@@ -0,0 +1,15 @@
+# Only for cartographer V4 with adxl
+# You need to include cartographer_survey.cfg and set the correct serial/canid in your overrides for this to work
+[adxl345] 
+cs_pin: scanner:ADXL_CS
```
TauNeutrino: I am just testing your PR- shouldn't that line be:
`cs_pin: scanner:MCU_ADXL`
since that is the alias defined in the mcu_definitions? or change alias the definitions - i dont know the correct naming convention 
I seem to run into a problem during the config because of that

config/hardware/probes/cartographer_survey.cfg
```diff
@@ -0,0 +1,33 @@
+# This probe type is for the Cartographer 3D probe used directly as a virtual Z endstop
+# rather than with an existing physical endstop. To use this configuration,
+# you will need to manually add the Cartographer 3D Klipper plugin!
+
+## Inside your mcu.cfg file, you need to add the cartographer as mcu like this:
```
TauNeutrino: since the board pin alias definitions file is not mentioned anywhere it would be nice to have it at least mentioned here in the example.
`[include config/mcu_definitions/scanner/Cartographer3D_V3.cfg] #for the V3 board` 
`[include config/mcu_definitions/scanner/Cartographer3D_V4.cfg] #for the V4 board` 

without them the correct ADXL pin cannot be determined by klipper

config/hardware/accelerometers/cartographer_adxl_accelerometer.cfg
TauNeutrino, line 4: I ran into problems with this because the alias in the mcu_definitions is `MCU_ADXL`
i dont know the proper naming convention but either this is corrected to `MCU_ADXL` or the alias definitions are corrected in order for this to work out of the box

ErdbeermarmeladebrotmitHonig: I think this naming convention is ok, at least is it done similar in other ADXL configs. And with my new commit from today, the assignment of these pins is done in the user templates I just added. So this is not an issue in my opinion.

ErdbeermarmeladebrotmitHonig:
Hi @Frix-x ,
this PR is now open for quite some time now. Are you going to merge it or do you want to have something additional in order to merge it?
Best Regards
Michael

DisDis:
It works for Cartographer 3D V3

jpiccari:
https://github.com/Frix-x/klippain/blob/develop/config/hardware/probes/cartographer_touch.cfg

Looks like we may soon get an official carto v3 and v4 setup.

mjonuschat:
Implemented in the Z probe framework, confirmed working after 6ea2917db10957f2068b6ac7b7b133ef2c05d36d.

Closing this out as duplicate.

PR #778, opened 2026-06-11
Align Cartographer workflow with updated home and print settings

ubrider:
- Remove hooks for cartographer and define it like a virtual z probe (like beacon virtual).
- Create a module and gcode macro for Cartographer_touch_home. (it's all it needs for touch). And modify start_print sequence according to cartographer documentation.
- Syntax/typo fixes.


mjonuschat:
Thanks for the PR - I used a slight variation of it in 6ea2917db10957f2068b6ac7b7b133ef2c05d36d to fix cartographer support.

PR #816, opened 2026-09-09
fix(mcu): Fix several ebb36 gen2 configuration defaults

anunknownsource:
I noticed several minor corrections or nice-to-haves for the BTT EBB36 Gen 2 board templates. This includes pin definitions for the SPI bus in the mcu_definitions folder (pins assigned to wrong definition), adding the pin remapping to the user_templates cfg, correcting the axis map to the most common orientation (to avoid an unnecessary override), and adding a commented serial line to the user_templates cfg for connections over USB vice canbus.


mjonuschat:
Thanks for the fix, this looks correct for EBB36 GEN2. One thing while you're in here:

`config/mcu_definitions/toolhead/BTT_EBB42_GEN2_v1.0.cfg` line 21 still has the old SPI2 mapping (SCLK=PB2, MISO=PB11, MOSI=PB10). EBB36 and EBB42 GEN2 share the same accelerometer config and the same spi2_PB2_PB11_PB10 bus, so it looks like this needs the same swap as the EBB36 fix.

anunknownsource:
I can make the same adjustments to the EBB42 later tonight. Not sure about the axes map, but will look at the diagrams and see if the chip is oriented similarly to the 36.

anunknownsource:
Corrected the EBB42 Gen 2 similarly to the EBB36 Gen 2, but have not looked at the axes map yet.

anunknownsource:
Looking at an old commit to add the ebb42, the suggested override was the same x,-z,y orientation, and it uses the same cfg file under accelerometers anyway, so it looks like the commit for the 36 also does the same for the 42. Also seems the orientation is similar for both if mounted in the "usual" way.

anunknownsource:
Do I need to open an issue for this or is the pull request how you want this done? Looking at old commits and pull requests it seems there's a mix bag of either method... What is the team's preference so I know what to do in the future?

